PR #13, opened 2022-02-28
Issue57

typelogic:


sonar-project.properties
```diff
@@ -0,0 +1,2 @@
+sonar.exclusions=.github/**, .vscode/**, android/**, assets/**, build/**, ios/**, node_modules/**, scripts/**
```
jeremi, line 1: @typelogic why this modification as part of this PR?

jeremi (changes requested):

sonar-project.properties
typelogic: i didn't intend to change this one.. i'll clean this up...

PR #33, opened 2022-06-08
Fixed app crashing and language issue

MonobikashDas:


MonobikashDas:
If multiple language is present for a field, then I am always taking the first one.

PR #80, opened 2022-08-24
[DSD-1364] Added delete workflow in manual mode for inji repo.

gaganamadival:


ckm007 (changes requested):
We dont need any input as such. Ps remove the variable input.

ckm007 (changes requested):
Is this tested in local

ckm007 (changes requested):
Also can you add the same workflow in the master branch as well and raise the PR.

ckm007 (commented):
merging as PR for default branch is merged.

PR #88, opened 2022-09-02
[DSD-1364] Added token to the workflow for inji repo.

gaganamadival:


ckm007 (changes requested):
Remove space between secrets. and ACTION_PAT

ckm007 (approved):

PR #392, opened 2023-01-06
Issue 212

Anil-kumar-Majji:


Anil-kumar-Majji:
Since the SingleVcItem.tsx was removed, there is no more requirement of resolving this issue.

PR #429, opened 2023-01-16
Qr login and wallet binding

MonobikashDas:


components/EditableListItem.tsx
```diff
@@ -48,5 +55,6 @@ export const EditableListItem: React.FC<EditableListItemProps> = (props) => {
 interface EditableListItemProps {
   label: string;
   value: string;
+  Icon: string;
```
srikanth716, line 58: Why are we using Icon in props?

PR #472, opened 2023-01-24
Corrected VC detailed layout ocupying the spaces properly

Anil-kumar-Majji:


Anil-kumar-Majji:
As per the new UI of the Vc Detailed page, the order of the contents and their headings are in different ways with spaces. So this issue wouldn't replicate now.

Issue #556, opened 2023-02-10
Error message : JSON Parse error: Unexpected EOF should be revised in inji app


Anushree09-N:
**Describe the bug**
We get  JSON Parse error: Unexpected EOF error when depend services are down or if there are issues in Khazana library.

**To Reproduce**
Was able to replicate the issue here https://github.com/mosip/inji/issues/554

**Current behavior**
Displayed error message " JSON Parse error: Unexpected EOF", when depend services are down or if there are issues in Khazana library.

**Expected behavior**
Error message should be revised and should be in user understandable form..

**Screenshots**

https://user-images.githubusercontent.com/79500509/218095832-afc9437d-f8d2-42ad-9a77-d3aa65c9b35e.mp4

![image](https://user-images.githubusercontent.com/79500509/218096277-0fb313de-e6c5-417f-81e9-53ee157929d2.png)

**Smartphone (please complete the following information):**

This is not device specific issue it's common across all

**Additional context**

https://github.com/mosip/inji/issues/520

damodarguru:
Verified this issue in qa-1201-b2 environment, still issue is occurring. We not getting appropriate error message in Arabic language if services are down or if there are any issues. I am attaching screenshot of device and the device details below for reference. Hence reopening the bug.

Screenshot: -
![image](https://user-images.githubusercontent.com/124761463/227144803-5801454d-6138-49d6-be26-2683b18ea038.png)
![image](https://user-images.githubusercontent.com/124761463/227144870-2af47857-1feb-4e61-9ecf-7eff4faaa9d0.png)

Tested in: -
Device: Redmi 7A
Android Version: 10
BLE version :4.2
Inji app link: https://github.com/mosip/inji/actions/runs/4447486079
Mimoto version: 1.2.0.1
MOSIP Version: qa-1201-b2.mosip.net

PR #561, opened 2023-02-14
#556 : URI change to esignet + error messages

MonobikashDas:


ravikp (approved):
Apart from the error messages, the code logic looks good to me.

Issue #600, opened 2023-03-01
Android - Unable to save VC 


santhoshsunder:
**Describe the bug**
Unable to save VC, save ID button I not responsive for a while 

**To Reproduce**
Prerequiste: VC is downloaded and stored in sharing device
Device A - verifier device
Device B - wallet device

steps:
1. open qr code in Device A
2. open scanner in Device B
3. scan the QR code from Device B
4. Perform a sharing method
5. Click on save id button on the device A

**Expected behavior**
The button should be responsive and VC should be saved in the single click

**Actual behavior**
The "save id" button is not responsive, its taking lot of time and saving intermittently or stuck on the that itself 

**Video :**
https://user-images.githubusercontent.com/102220709/222070783-aedd67ad-8cd8-41d9-8178-11e22e0f43ba.mp4

**Add the screenshot of the profile page with commit id**
https://user-images.githubusercontent.com/102220709/222070879-dac77bb9-0787-49e1-871e-26e79e51839d.mp4

**Smartphone :**

Wallet Device: Device B
  - Phone make/model: Redmi Note 10 lite
  - OS: Android 10
  - BLE version : 5.0

Verifier Device: Device A
  - Phone make/model: Vivo Y73
  - OS: Android 12
  - BLE version : 5.0

MOSIP Version : 1.2.0.1
MOSIP server : qa-1201-b2.net
APP link : https://github.com/mosip/inji/actions/runs/4293687772

ravikp:
Looks like the APK link points to an older build. Kindly test with latest build. Also the issue is related to having large verifiable credentials downloaded as part of the previous transfers.

santhoshsunder:
Verified the bug with the size less than 30 MB, below are the observations:

1. Save button is getting disabled after first click.
2. Able to share multiple unique VCs and save the same in the Verifier phone without any hassle.
3. Able to download VC too.

When the size crosses 30 MB, we are facing issues which will be tracked in the issue #625
Hence closing the bug.

PR #602, opened 2023-03-02
fix(#600): Disable accept and reject button once clicked

ravikp:
Once the button is clicked, the save and discard buttons needed to be disabled.


ravikp:
This is coming as part of the release. Hence closing the ticket.

PR #693, opened 2023-03-29
0.9

MonobikashDas:


ravikp (approved):
Looks good to me, as this fixes the application being hung when there is a failure in face authentication. This is the reason for the upgrade.

Issue #696, opened 2023-03-30
Verify Inji build on windows machine


krishnakumar4a4:
Verify if Inji can be built seamlessly on Windows machine as a developer

Pragya279:
This bugs has been moved to Jira.
Link to Jira bug: https://mosip.atlassian.net/browse/MOSIP-26894

PR #712, opened 2023-04-17
chore(#696): Fix compile errors on windows machine due to Tuvali's incorrect build scripts for windows machine

Alka1703:
This fixes the compile time errors occurring in Inji due to Tuvali's npm executable scripts. These scripts were originally build for mac machine in the initial days when Inji app was not there. Now these Tuvali executable scripts aren't used.


ravikp (approved):
Looks good for me.

PR #737, opened 2023-06-02
iOS Version and fastlane changes for New ui

dhivya0413:


ravikp (approved):
Looks good to me. For the new UI, we are bumping up the version from `0.9.x` to `0.10.x`

PR #747, opened 2023-06-16
 INJI-31 -  [Android] Added a workflow to publish to play store

adityankannan-tw:
[ Android ]

-  Added a workflow to publish to play store internal testing track.
- Updated the read me on keystore changes.

When merged, we will be able to trigger the android custom build and pipeline will build the app and push the app bundle to Internal testers track. Then the app download link should be shared with the testers whose email id's are registered in google play console internal testers track. The testers will be able to download the apk from the play store with that link.

NOTE :  Testers will be able to access the link and download the apk only when their email id's are registered in the Internal testing track in google play console.


ckm007 (approved):

README.md
```diff
@@ -15,7 +15,23 @@ Be sure to have the following build tools installed before proceeding:
...
+ -genkey -v \
+ -storetype PKCS12 \
+ -keyalg RSA \
+ -keysize 2048 \
+ -validity 10000 \
+ -storepass 'android' \
+ -keypass 'android' \
+ -alias androiddebugkey \
+ -keystore android/app/debug.keystore \
+ -dname "CN=io.mosip.residentapp,OU=,O=,L=,S=,C=US"
```
krishnakumar4a4, line 29: For release, find out the right registration details give for mosip domain and update.

adityankannan-tw: Since it has been created already and used for the first time. If we want to change the details we need to delete the app from play store and configure it again in the play console. How to proceed with this?

adityankannan-tw: We have re-configured the app with the keystore details and updated the same in the secrets.

PR #784, opened 2023-08-30
Inji 304 update tuvali and secure keystore version

KiruthikaJeyashankar:


shared/constants.ts
```diff
@@ -33,13 +33,6 @@ export const isSameVC = (vcKey: string, pinnedVcKey: string) => {
   return requestId === pinnedRequestId;
 };
 
-//To compare the vckey with requestId, when the vc is pinned
-export const isSameVC = (vcKey: string, pinnedVcKey: string) => {
```
swatigoel, line 37: why removing this method?

swatigoel: got it. added twice

swatigoel (approved):

shared/constants.ts
KiruthikaJeyashankar: This method is declared twice with same content. Thus making build fail
Removed this to avoid build failure @swatigoel

PR #789, opened 2023-08-31
fix(INJI-247):The VC's three dots elipsise layout is not proper

srikanth716:
IOS :
IMG_5203

Android:
Screenshot_1693487006


shared/constants.ts
```diff
@@ -33,13 +33,6 @@ export const isSameVC = (vcKey: string, pinnedVcKey: string) => {
   return requestId === pinnedRequestId;
 };
 
-//To compare the vckey with requestId, when the vc is pinned
```
swatigoel, line 36: It's not needed. Please take latest pull from develop first and then raise PR

swatigoel (changes requested):

swatigoel (approved):

PR #793, opened 2023-09-01
fix(INJI-268): Adjusted alignment of page headers properly

Anil-kumar-Majji:
**[INJI-268](https://mosip.atlassian.net/browse/INJI-268)** - Page headers are not aligned properly 

[INJI-268]: https://mosip.atlassian.net/browse/INJI-268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/KebabPopUp.tsx
+++ b/components/KebabPopUp.tsx
@@ -3,13 +3,14 @@ import { Icon, ListItem, Overlay } from 'react-native-elements';
 import { Theme } from '../components/ui/styleUtils';
 import { Column, Row, Text } from '../components/ui';
 import { WalletBinding } from '../screens/Home/MyVcs/WalletBinding';
-import { Pressable } from 'react-native';
+import { Pressable, View } from 'react-native';
 import { useKebabPopUp } from './KebabPopUpController';
 import { ActorRefFrom } from 'xstate';
 import { vcItemMachine } from '../machines/vcItem';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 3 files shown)


components/KebabPopUp.tsx
```diff
@@ -26,6 +27,7 @@ export const KebabPopUp: React.FC<KebabPopUpProps> = (props) => {
         onBackdropPress={props.onDismiss}
         overlayStyle={Theme.KebabPopUpStyles.kebabPopUp}>
         <Row style={Theme.KebabPopUpStyles.kebabHeaderStyle}>
+          <View></View>
```
swatigoel, line 30: Why this empty view?

swatigoel (changes requested):

components/KebabPopUp.tsx
Anil-kumar-Majji: Here I used **View, Text** and **Icon** in the row. Then I can able to separate the elements with 
"**justify-content: space-between**". This aligns the header elements of the kebab pop-up.

swatigoel: Why we need empty view to make header in the center. We should update text style to make it center.
For kebab popup, I don't see any icon in header coming in near future.

swatigoel (approved):

PR #803, opened 2023-09-11
testIds for intro and language screens

PoojaBabusing:
PR for issue: https://mosip.atlassian.net/browse/INJI-335

### DIFF
--- a/components/ui/SetupPicker.tsx
+++ b/components/ui/SetupPicker.tsx
@@ -29,6 +29,7 @@ export const SetupPicker: Picker = (props: PickerProps<unknown>) => {
 
   return (
     <Column
+      testID={props.testID}
       width={Dimensions.get('window').width * 0.8}
       backgroundColor={Theme.Colors.whiteBackgroundColor}>
       {props.items.map((item, index) => (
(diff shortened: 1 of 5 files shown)


screens/Home/IntroSlidersScreen.tsx
```diff
@@ -80,14 +82,19 @@ export const IntroSlidersScreen: React.FC<RootRouteProps> = (props) => {
           </Row>
           <Image source={item.image} />
           <Column
+            testID="introSlideColumn"
```
swatigoel, line 85: can we name it better?
"introSlide" will be better instead of "introSlideColumn"

components/ui/SetupPicker.tsx
```diff
@@ -29,6 +29,7 @@ export const SetupPicker: Picker = (props: PickerProps<unknown>) => {
 
   return (
     <Column
+      testID={props.testID}
```
swatigoel, line 32: This is the wrapper of rendering items like list of languages.
Need to add testId to each list item and that has to be unique for each item.

swatigoel (changes requested):

screens/Home/IntroSlidersScreen.tsx
PoojaBabusing: done

components/ui/SetupPicker.tsx
PoojaBabusing: refactored this

swatigoel (approved):

PR #833, opened 2023-09-13
add testIds for pincard and vc activation

PoojaBabusing:


swatigoel:
@PoojaBabusing can we use testID as "warningMsg" for binding and removing? Other testID are common in both the overlay so that it will help QA to automate as single locator

swatigoel (approved):

PR #837, opened 2023-09-18
Add support to change mimoto and esignet host url under credential registry

swatigoel:
[INJI 360](https://mosip.atlassian.net/browse/INJI-360) - Add support to change mimoto and esignet host url under credential registry.


screens/Profile/ProfileScreen.tsx
```diff
@@ -58,20 +58,32 @@ export const ProfileScreen: React.FC<MainRouteProps> = (props) => {
           title={controller.alertMsg}
         />
         <EditableListItem
-          label={t('name')}
-          value={controller.name}
+          title={t('name')}
```
vijay151096, line 61: Profile Screen is unused, it is stale code in inji, no need any changes here, as we should be removing this file

screens/Profile/ProfileScreenController.ts
```diff
@@ -110,16 +110,14 @@ export function useProfileScreen({ navigation }: MainRouteProps) {
     canUseBiometrics: useSelector(authService, selectCanUseBiometrics),
     useBiometrics,
 
-    UPDATE_NAME: (name: string) =>
-      settingsService.send(SettingsEvents.UPDATE_NAME(name)),
+    UPDATE_NAME: (names) =>
```
vijay151096, line 113: "Profile Screen is unused" - please check and remove if not needed.

screens/Profile/ProfileScreen.tsx
swatigoel: I haven't removed Profile related code with assumption that we might leverage this structure with upcoming features.
Made changes so that it aligns with current implementation

screens/Profile/ProfileScreenController.ts
swatigoel: I haven't removed Profile related code with assumption that we might leverage this structure with upcoming features.
Made changes so that it aligns with current implementation

vijay151096 (approved):
looks good to me.

PR #839, opened 2023-09-19
fix(inji-324): show the download popup until the card is loaded successfully in the UI

PuBHARGAVI:


screens/Home/MyVcsTabController.ts
```diff
@@ -56,6 +57,10 @@ export function useMyVcsTab(props: HomeScreenTabProps) {
       settingsService,
       selectShowHardwareKeystoreNotExistsAlert,
     ),
+    isVcLoaded: useSelector(vcService, selectIsVcLoaded),
```
tilak-puli, line 60: can we renamed these variables from Loaded to downloaded?

loading can be from storage also.

swatigoel (approved):

PR #854, opened 2023-09-22
feat(INJI-343): Text changes & other UI enhancements

vharsh:
### MessageOverlay component

- [UI] gets another button & corresponding action with a configurable text element in **MessageOverlayProps** ( for popups with texts like `ok, cancel, try-again, abort, delete` i.e. the button text becomes more generic )
- [UI] gets a custom `height` property to update the stylesheet, it gives other usage locations more control to adjust the sizes without breaking them or fixing the bugs for them(i.e. no hidden changes are expected from this change)

## User Visible changes

**Before**:  (please ignore phone time as the bug has been simulated to take screenshots)
Screenshot 2023-09-22 at 4 50 50 PM

**After**:

Screenshot 2023-09-22 at 4 48 32 PM

ref: https://mosip.atlassian.net/browse/INJI-343


components/DualMessageOverlay.tsx
```diff
@@ -1,21 +1,29 @@
...
-) => {
-  const { t } = useTranslation('common');
+export const DualMessageOverlay: React.FC<DualMessageOverlayProps> = props => {
+  const {t} = useTranslation('common');
+  const style = StyleSheet.create({
+    customHeight: {
+      height: props.customHeight
+        ? props.customHeight
+        : props.progress
+        ? 100
```
swatigoel, line 22: Why can't we have it from theme as it already exists `Theme.MessageOverlayStyles.popupOverLay`
And why we want to create this as style instead of just height?

components/MessageOverlay.tsx
```diff
@@ -1,13 +1,22 @@
...
 export const MessageOverlay: React.FC<MessageOverlayProps> = props => {
   const {t} = useTranslation('common');
+  const style = StyleSheet.create({
+    customHeight: {
+      height: props.customHeight
+        ? props.customHeight
+        : props.progress
+        ? 100
+        : 150,
+    },
```
swatigoel, line 18: Same comment as in DualMessageOverlay

components/MessageOverlay.tsx
```diff
@@ -60,6 +65,14 @@ export const MessageOverlay: React.FC<MessageOverlayProps> = props => {
             styles={Theme.MessageOverlayStyles.button}
           />
         ) : null}
+        {!props.children && props.onCustomBtnTxt ? (
```
swatigoel, line 68: If we pass onCancel and onCustomBtnTxt both, are we expecting to render both the buttons?
If yes, can you please attach screenshot how does it look?
if no, then we have to make change that it MessageOverlay renders only one button, the text and action of the button is received as props

screens/Issuers/IssuersScreen.tsx
```diff
@@ -132,7 +132,7 @@ export const IssuersScreen: React.FC<
           title={t(`errors.${controller.errorMessage}.title`)}
           message={t(`errors.${controller.errorMessage}.message`)}
           goBack={goBack}
-          tryAgain={isGenericError() ? null : controller.TRY_AGAIN}
+          tryAgain={controller.TRY_AGAIN}
```
swatigoel, line 135: @KiruthikaJeyashankar Please review this change

KiruthikaJeyashankar: Only for no internetConnection error we allow users to try again as per the UI wireframes. Are we having any usecase change here @vharsh ? If no this change shall be reverted

KiruthikaJeyashankar: Confirming from the wireframe in INJI-343, both generic error as well noInternetConnectionError has try again feature.

vharsh: As per the Jira description both the types of errors need to have the try-again error button, hence removed the check.

components/DualMessageOverlay.tsx
vharsh: Because we want to conditionally set the height property as per the set properties, hence I had to move the height values from the theme stylesheets to here.

screens/Scan/ScanLayoutController.ts
```diff
@@ -114,7 +114,7 @@ export function useScanLayout() {
     statusOverlay = {
       title: t('status.sharingInProgress'),
       hint: t('status.connectingTimeout'),
-      onCancel,
+      onButtonPress: onCancel,
```
swatigoel, line 117: At line no 99, onCancel should be replaced by onButtonPress

swatigoel (changes requested):

swatigoel (approved):

PR #859, opened 2023-09-26
upload app file to actions

adityankannan-tw:


package.json
```diff
@@ -85,7 +85,7 @@
     "react-native-spinkit": "^1.5.1",
     "react-native-svg": "13.4.0",
     "react-native-swipe-gestures": "^1.0.5",
-    "react-native-tuvali": "github:mosip/tuvali#v0.4.4",
+    "react-native-tuvali": "github:tw-mosip/tuvali#tuvali-ios-error",
```
swatigoel, line 88: why tw-mosip?

adityankannan-tw: So building for simulator it fails with the error, where 1 method in tuvali code ( ceill(totalChunkCount) ) it is containing 2 l's and it is failing. So removed that and it was successful. Now need to check with physical device also, but my phn got updated to ios17 now xcode also being updated to work with 17.

adityankannan-tw: In chunker.swift file

swatigoel (approved):

PR #862, opened 2023-09-26
Add a new workflow for ios automation build

adityankannan-tw:


swatigoel:
@ckm007 We have to create .app file in iOS for UI automation. So we want to have it as separate workflow.

dhivya0413 (approved):
LGTM

PR #864, opened 2023-09-26
fix(Inji 332): reset UIN field in add vc modal and set AddVcModal dropdown to VID

PuBHARGAVI:
- Reset the UIN field of AddVcModal after getting the UIN/VID using AID and downloading the VC
- Set the default value of AddVcModal dropdown to VID as Resident services is returning VID when we try to fetch UIN/VID using AID


screens/Home/MyVcs/GetVcModalMachine.ts
```diff
@@ -318,7 +319,10 @@ export const GetVcModalMachine =
               transactionID: context.transactionId,
             }
           );
-          return response.response.individualId;
+          return {
+            id: response.response.individualId,
+            idType: response.response.individualIdType || 'VID',
```
swatigoel, line 324: set default to UIN

swatigoel (changes requested):

screens/Home/MyVcs/GetVcModalMachine.ts
PuBHARGAVI: Done. changed it to UIN

swatigoel (approved):

PR #871, opened 2023-09-28
 feat(inji-344): Fix pin status not updating in vc item context

tilak-puli:


components/VC/ExistingMosipVCItem/ExistingMosipVCItem.tsx
```diff
@@ -1,4 +1,4 @@
-import React, {useContext, useEffect, useRef} from 'react';
+import React, {useContext, useEffect, useMemo, useRef} from 'react';
```
swatigoel, line 1: are we using useMemo?

swatigoel (approved):

PR #879, opened 2023-09-29
feat(INJI-415): Delete extra patch files.

swatigoel:


vijay151096 (approved):
looks good, as these are older patches.

PR #881, opened 2023-09-29
feat(INJI-410): update success message for QR login

swatigoel:


locales/en.json
```diff
@@ -330,7 +330,7 @@
     "domainWarning": "Please confirm the domain of the website you are scanning the QR code from below",
     "access": " is requesting access to",
     "status": "Status",
-    "successMessage": "You are successfully logged in ",
+    "successMessage": "You are successfully logged in to ",
```
vijay151096, line 333: do we need to update the literals for other languages as well ?

swatigoel: no, this is for demo only (in english)

vijay151096 (approved):
lgtm

PR #883, opened 2023-09-29
Inji-400: Cache api calls and return initial values if network not available

tilak-puli:


shared/InitialConfig.ts
```diff
@@ -0,0 +1,20 @@
+// Initial configs used by app if network is not available when app is opened for first time. once network is available response is cached and used from then
+// Note: Need to keep this config in sync with actual mimoto response.
```
swatigoel, line 2: add configuration file name here
In mosip-config, refer inji-default.properties

shared/InitialConfig.ts
```diff
@@ -0,0 +1,20 @@
...
+    audience: 'ida-binding',
+    allowedInternalAuthType: 'otp,bio-Finger,bio-Iris,bio-Face',
+    vcDownloadMaxRetry: '10',
+    minStorageRequiredForAuditEntry: '2',
+    minStorageRequired: '2',
+    vcDownloadPoolInterval: '6000',
+    issuer: 'residentapp',
+    allowedAuthType: 'demo,otp,bio-Finger,bio-Iris,bio-Face',
+    allowedEkycAuthType: 'demo,otp,bio-Finger,bio-Iris,bio-Face',
+    warningDomainName: 'https://api.qa-inji.mosip.net',
```
swatigoel, line 16: Remove warningDomainName and faceSdkModelUrl from caching as it depends on api response.
Let app handles these errors if these values are missing

shared/api.ts
```diff
@@ -0,0 +1,161 @@
...
+};
+
+interface cacheAPIMethodProps {
+  isCachePreferred?: boolean;
+  cacheKey: string;
+  fetchCall: (...props: any) => Promise<any>;
+  onErrorHardCodedValue?: any;
+}
+
+async function catchAPIMethod({
```
swatigoel, line 82: Do we need to use keyword `method` in function name?
and it's catchAPI or cacheAPI?

shared/api.ts
```diff
@@ -0,0 +1,161 @@
...
+      const response = await fetchCall();
+
+      Storage.setItem(cacheKey, JSON.stringify(response)).then(() =>
+        console.log('Cached response for ' + cacheKey),
+      );
+
+      return response;
+    }
+  } catch (error) {
+    console.warn('Failed to load due to network issue');
```
swatigoel, line 123: Add function name in warning message which will help to understand this error has come for caching api

shared/api.ts
```diff
@@ -0,0 +1,161 @@
...
+  } else {
+    return await cacheAPIMethodWithAPIPreference(
+      cacheKey,
+      fetchCall,
+      onErrorHardCodedValue,
+    );
+  }
+}
+
+async function cacheAPIMethodWithCachePreference(
```
swatigoel, line 103: Same here do we need to use keyword method in function name?

shared/api.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    if (onErrorHardCodedValue != undefined) {
+      return onErrorHardCodedValue;
+    } else {
+      throw error;
+    }
+  }
+}
+
+async function cacheAPIMethodWithAPIPreference(
```
swatigoel, line 134: Same here do we need to use keyword method in function name?

shared/api.ts
```diff
@@ -0,0 +1,161 @@
...
+  try {
+    const response = (await Storage.getItem(cacheKey)) as string;
+
+    if (response) {
+      return JSON.parse(response);
+    } else {
+      const response = await fetchCall();
+
+      Storage.setItem(cacheKey, JSON.stringify(response)).then(() =>
+        console.log('Cached response for ' + cacheKey),
```
swatigoel, line 117: This should be logged in debugging mode only

shared/api.ts
```diff
@@ -0,0 +1,161 @@
...
+
+async function cacheAPIMethodWithAPIPreference(
+  cacheKey: string,
+  fetchCall: (...props: any[]) => any,
+  onErrorHardCodedValue?: any,
+) {
+  try {
+    const response = await fetchCall();
+    Storage.setItem(cacheKey, JSON.stringify(response)).then(() =>
+      console.log('Cached response for ' + cacheKey),
```
swatigoel, line 142: This should be logged in debugging mode only

swatigoel (changes requested):

shared/api.ts
tw-tilak-puli: All logs are only visible in debug mode

shared/InitialConfig.ts
tilak-puli: sure. adding

shared/InitialConfig.ts
tilak-puli: removed

shared/api.ts
```diff
@@ -0,0 +1,161 @@
...
+  } else {
+    return await generateCacheAPIFunctionWithAPIPreference(
+      cacheKey,
+      fetchCall,
+      onErrorHardCodedValue,
+    );
+  }
+}
+
+async function generateCacheAPIFunctionWithCachePreference(
```
swatigoel, line 103: This is the actual method which use cached response or fetch data from api.
We can name it as cacheAPIWithCachePreference

swatigoel (changes requested):

shared/InitialConfig.ts
tilak-puli: removed now. after removing also app works fine

shared/InitialConfig.ts
tilak-puli: added near all properties key

swatigoel (approved):

PR #885, opened 2023-09-29
feat(INJI-414): Update copy text messages

swatigoel:
feat(INJI-414): Update copy text message for QR login flow and unlock application page


vijay151096 (approved):
looks good to me.

PR #903, opened 2023-10-10
On click of Download card, If the internet if Off Display no internet connection page 

PoojaBabusing:
This PR is for the issue: https://mosip.atlassian.net/browse/INJI-357


locales/en.json
```diff
@@ -299,6 +299,10 @@
         "title": "Some security features will be unavailable",
         "message": "Your current device does not support all the security features.",
         "riskOkayText": "Ok"
+      },
+      "noInternetConnection": {
+        "title": "No internet connection",
+        "message": "Please check your connection and retry"
```
swatigoel, line 305: Don't we need this in other language?

swatigoel (changes requested):

components/ui/Error.tsx
```diff
@@ -6,10 +6,38 @@ import {Button, Column, Row, Text} from '.';
 import {Header} from './Header';
 import {Theme} from './styleUtils';
 import testIDProps from '../../shared/commonUtil';
+import {Modal} from './Modal';
 
 export const Error: React.FC<ErrorProps> = props => {
```
Anil-kumar-Majji, line 11: Will we use this same component to show any other errors that might be in the future? If it's specific then set the component title to that particular error.

swatigoel (approved):

locales/en.json
PoojaBabusing: added the translations

PR #911, opened 2023-10-11
feat(INJI-419): notify user on minStorageLimitReached while vc download

KiruthikaJeyashankar:
When user clicks on plus icon in home page for downloading VC, if the device storage is less than the min threshold defined in our inji-default.properties file, error message overlay is shown (attached screenshot)

mosip.inji.minStorageRequired=5

image

### DIFF
--- a/screens/Home/HomeScreen.tsx
+++ b/screens/Home/HomeScreen.tsx
@@ -14,6 +14,7 @@ import {ExistingMosipVCItemMachine} from '../../machines/VCItemMachine/ExistingM
 import {isOpenId4VCIEnabled} from '../../shared/openId4VCI/Utils';
 import LinearGradient from 'react-native-linear-gradient';
 import {EsignetMosipVCItemMachine} from '../../machines/VCItemMachine/EsignetMosipVCItem/EsignetMosipVCItemMachine';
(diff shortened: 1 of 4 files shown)


swatigoel (approved):

screens/Home/HomeScreen.tsx
```diff
@@ -76,6 +77,12 @@ export const HomeScreen: React.FC<HomeRouteProps> = props => {
         )}
       </Column>
       {isOpenId4VCIEnabled() && <DownloadFABIcon />}
+      <ErrorMessageOverlay
```
vijay151096, line 80: Have we removed the overlay from MyVcsTab.tsx, which was used earlier in the downloadcard scenario

KiruthikaJeyashankar: We have not cleaned up the download card button still so I have not removed this storage error overlay as well from MyVCsTab.tsx

vijay151096: okay, can you please add this as a sub task in the same clean up story

KiruthikaJeyashankar: Noted

PR #915, opened 2023-10-11
talisman check for pre-commit and pre-push hooks

PoojaBabusing:
This PR is for issue: https://mosip.atlassian.net/browse/INJI-413

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "i18n:compile-strings": "node scripts/compile-strings.js",
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx --fix",
     "test": "jest",
-    "postinstall": "patch-package && yarn jetify && sh tools/talisman/talisman-precommit.sh"
+    "postinstall": "patch-package && npm run jetify && sh tools/talisman/talisman-precommit.sh"
   },
   "dependencies": {
     "@digitalbazaar/ed25519-signature-2018": "digitalbazaar/ed25519-signature-2018",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "scripts": {
     "prepare": "husky install",
+    "jetify": "jetify",
     "start": "react-native start",
     "android:mosip": "react-native run-android --variant=mosipDebug",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,14 +1,15 @@
...
+    "jetify": "jetify",
     "start": "react-native start",
     "android:mosip": "react-native run-android --variant=mosipDebug",
     "ios": "react-native run-ios",
     "build:android:mosip": "cd android && ./gradlew :app:assembleMosipRelease && cd ..",
     "i18n:compile-strings": "node scripts/compile-strings.js",
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx --fix",
     "test": "jest",
-    "postinstall": "patch-package"
+    "postinstall": "patch-package && yarn jetify && sh tools/talisman/talisman-precommit.sh"
```
tilak-puli, line 12: can we use npm instead of yarn as we use npm by default

tools/talisman/talisman-precommit.sh
```diff
@@ -0,0 +1,35 @@
...
+    fi
+
+    if [ -f $prePushFile ] && [ -z $GITHUB_ACTIONS ]
+    then
+        rm $prePushFile
+    fi
+
+    if [ ! -f $preCommitFile ] && [ -z $GITHUB_ACTIONS ]
+    then
+        curl https://thoughtworks.github.io/talisman/install.sh > ~/install-talisman.sh
```
tilak-puli, line 22: @swatigoel we are downloading talisman install file from thoughtworks github.

this should fine right? if we don't want to do this then we have store shell script in our repo

package.json
PoojaBabusing: Done replaced with npm run jetify

tools/talisman/talisman-precommit.sh
swatigoel: Let it be downloading from thoughtworks github. It's open source, so should not be worry

tilak-puli (approved):

swatigoel (approved):

PR #916, opened 2023-10-11
fix(INJI-205): vc download via esignet

KiruthikaJeyashankar:
remove conversion of json in downloaded credential response as request method called is already returning json type response. This eliminates error while vc download via eSignet


vijay151096 (approved):
changes looks good to me.

PR #933, opened 2023-10-17
[WIP] feat(INJI-438): show consent based on api response in QR login flow

KiruthikaJeyashankar:
In the QR login flow the consent screen will be displayed as per the response from esignet /linked-authorization/v2/authenticate.

- When a binded VC is used in QR login for the first time the consent screen is shown in Inji( this is stored in eSignet side for further login)
- When a binded VC is used in QR login for the second (2+n) time where consent was provided by the user in the first login itself, Consent screen is not shown to the user (as /linked-authorization/v2/authenticate returns "NOCAPTURE")


components/ui/themes/DefaultTheme.ts
```diff
@@ -1298,14 +1299,14 @@ export const DefaultTheme = {
...
-  if (Platform.OS === 'ios') {
+  if (isIOS()) {
     return {
       shadowColor: '#000',
       shadowOffset: {width: 1, height: 1.2},
       shadowOpacity: 0.3,
       shadowRadius: 2.5,
     };
-  } else if (Platform.OS === 'android') {
+  } else {
```
swatigoel, line 1309: we can remove else and directly return
so it will be 
```
if(isIOS()) {
return a;
}
return b;
```

components/ui/themes/PurpleTheme.ts
```diff
@@ -1295,14 +1296,14 @@ export const PurpleTheme = {
...
-  if (Platform.OS === 'ios') {
+  if (isIOS()) {
     return {
       shadowColor: '#000',
       shadowOffset: {width: 1, height: 1.2},
       shadowOpacity: 0.3,
       shadowRadius: 2.5,
     };
-  } else if (Platform.OS === 'android') {
+  } else {
```
swatigoel, line 1306: same here, we can remove else

machines/QrLoginMachine.ts
```diff
@@ -162,6 +162,19 @@ export const qrLoginMachine =
               target: 'showvcList',
             },
           },
+          initial: 'idle',
```
swatigoel, line 165: why we need idle as empty state here?

swatigoel (changes requested):

components/ui/themes/PurpleTheme.ts
swatigoel: ignore this, I have handled this in tech debt story

components/ui/themes/DefaultTheme.ts
swatigoel: ignore this, I have handled this in tech debt story

PR #949, opened 2023-10-19
feat(Inji-435): add telemetry events for vc activation flow

PuBHARGAVI:
-Added Telemetry events for the VC activation flow
-Increased the Pressable area of kebab icon in VC


components/KebabPopUpController.tsx
```diff
@@ -94,6 +99,7 @@ export function useKebabPopUp(props) {
     );
     otpError = useSelector(service, esignetSelectOtpError);
     walletBindingError = useSelector(service, esignetSelectWalletBindingError);
+    bindingAuthFailedError = useSelector(service, selectBindingAuthFailedError);
```
swatigoel, line 102: It's redundant

components/KebabPopUpController.tsx
```diff
@@ -66,6 +67,10 @@ export function useKebabPopUp(props) {
   let isWalletBindingError = useSelector(service, selectShowWalletBindingError);
   let otpError = useSelector(service, selectOtpError);
   let walletBindingError = useSelector(service, selectWalletBindingError);
+  let bindingAuthFailedError = useSelector(
```
swatigoel, line 70: It can const as it's not changed later

machines/VCItemMachine/EsignetMosipVCItem/EsignetMosipVCItemMachine.ts
```diff
@@ -32,10 +41,8 @@ const model = createModel(
     verifiableCredential: null as VerifiableCredential,
     isPinned: false,
     hashedId: '',
-
     publicKey: '',
     privateKey: '',
-
```
swatigoel, line 38: Looks like, by mistake deleted these

shared/telemetry/TelemetryUtils.js
```diff
@@ -135,27 +135,41 @@ export function getAppInfoEventData() {
   };
 }
 
-let passcodeRetryCount = 1;
+let retryCount = 1;
```
swatigoel, line 138: we should set retryCount as 0 by default

shared/telemetry/TelemetryUtils.js
```diff
@@ -135,27 +135,41 @@ export function getAppInfoEventData() {
...
+            TelemetryConstants.ErrorId.resend,
+            TelemetryConstants.ErrorMessage.resendOtp,
+          ];
+    sendErrorEvent(getErrorEventData(eventType, errorId, errorMessage));
+    retryCount = 1;
   }
 };
 
+export const resetRetryCount = () => {
+  retryCount = 1;
```
swatigoel, line 160: here also retryCount = 0

swatigoel (changes requested):

machines/VCItemMachine/EsignetMosipVCItem/EsignetMosipVCItemMachine.ts
PuBHARGAVI: @swatigoel empty lines?

shared/telemetry/TelemetryUtils.js
PuBHARGAVI: done

shared/telemetry/TelemetryUtils.js
PuBHARGAVI: done

components/KebabPopUpController.tsx
PuBHARGAVI: @swatigoel we are replacing it with esignet machine function if we are using openId4VCI. So we have to keep it as let

PuBHARGAVI: @swatigoel i think we don't need this if condition for openI4VCI becuase in props we are already getting exisitngMosipVcItem machine or EsignetMosipVcItemMachine as props and when we call the functions it will call the corresponding machine functions only so we don't need this if condition. if there is any function which is present in one machine but not in other we can handle those functions in if condition. I will check with @vijay151096 regarding this as he added the code

PuBHARGAVI: I have extracted the common functions of EsignetMosipVcItemMachine & exisitngMosipVcItemMachine into common file and removed if condition from kebabPopupController

components/KebabPopUpController.tsx
PuBHARGAVI: I have removed this as part of refactoring

machines/VCItemMachine/EsignetMosipVCItem/EsignetMosipVCItemMachine.ts
swatigoel: My bad, somehow I read it you removed publicKey and privateKey variables

components/KebabPopUpController.tsx
swatigoel: Validate this change once with @vijay151096

components/VC/MosipVCItem/VcItemController.tsx
```diff
@@ -50,19 +47,13 @@ export function useVcItemController(
...
   let generatedOn = useSelector(service, selectGeneratedOn);
   if (props.vcMetadata.isFromOpenId4VCI()) {
-    context = useSelector(service, esignetSelectContext);
-    isKebabPopUp = useSelector(service, esignetSelectKebabPopUp);
-    generatedOn = useSelector(service, esignetSelectGeneratedOn);
-    emptyWalletBindingId = useSelector(
-      service,
-      esignetSelectEmptyWalletBindingId,
-    );
+    context = useSelector(service, selectContext);
```
KiruthikaJeyashankar, line 50: This looks like duplicate assignment as the same is done in line:40

PuBHARGAVI: true didn't notice that. Thanks for pointing it out

KiruthikaJeyashankar (approved):

vijay151096 (approved):
looks good to me.

components/KebabPopUpController.tsx
vijay151096: validated.

swatigoel (approved):

PR #952, opened 2023-10-20
Refactor(inji-472): refactoring api usage, fix warnings, and fix fabicon ui issue

vijay151096:
**Issues Fixed as a part of this PR :**

1.  Make the FAB Icon pressable in all places. 
2. Masking the UIN/VID in the VC Details Page.
3. Introduced Padding at the bottom of the MyVCs page to make the FAB pressable in all device screen sizes. 
4. Moved all the API's to api.ts irrespective of cached.
5. Fixed IOS build issue for taking the inputted host endpoints.
6. Removed the package.json entry for telemetry-sdk


vharsh (approved):
lgtm

components/VC/MosipVCItem/MosipVCItem.tsx
```diff
@@ -73,7 +73,6 @@ export const MosipVCItem: React.FC<
               vcMetadata={props.vcMetadata}
               verifiableCredential={verifiableCredential}
               emptyWalletBindingId={emptyWalletBindingId}
-              onPress={() => props.onPress(service)}
```
swatigoel, line 76: why this is removed?

components/VC/MosipVCItem/MosipVCItemDetails.tsx
```diff
@@ -167,7 +168,7 @@ export const MosipVCItemDetails: React.FC<
                       weight="semibold"
                       size="smaller"
                       color={Theme.Colors.Details}>
-                      {uin}
+                      {getMaskedText(uin)}
```
swatigoel, line 171: In details, are we expecting to show masked data?

screens/Home/HomeScreen.tsx
```diff
@@ -44,15 +45,15 @@ export const HomeScreen: React.FC<HomeRouteProps> = props => {
       <LinearGradient
         colors={Theme.Colors.gradientBtn}
         style={Theme.Styles.downloadFabIcon}>
-        <Button
-          testID="downloadIcon"
-          icon={plusIcon}
+        <Pressable
```
swatigoel, line 48: We need to add testId to Fab icon.
As we are wrapping this in Pressable, just check whether we get it in appium for iOS and Android both.
If not, we might need to add `accessible={false}`

swatigoel (changes requested):

components/ui/themes/DefaultTheme.ts
```diff
@@ -529,14 +529,28 @@ export const DefaultTheme = {
...
       padding: 10,
-      backgroundColor: Colors.Orange,
       shadowColor: '#000',
       shadowOpacity: 0.4,
       elevation: 5,
       position: 'absolute',
       bottom: Dimensions.get('window').width * 0.1,
       right: Dimensions.get('window').width * 0.1,
     },
+    fabIconNormal: {
```
KiruthikaJeyashankar, line 539: Can we download  / plus word also here to be more explicit about this icon use-case (download card button)? like downloadFabIconNormal

KiruthikaJeyashankar (approved):

components/VC/MosipVCItem/MosipVCItemDetails.tsx
vijay151096: checked with @kratitva, new ui is coming up which will involve lot of ui refactoring, so for now, lets go with the existing approach. so removing this.

components/VC/MosipVCItem/MosipVCItem.tsx
vijay151096: onPress is not being used anywhere in the activation, Hence removing it.

components/ui/themes/DefaultTheme.ts
vijay151096: Updated it.

screens/Home/HomeScreen.tsx
vijay151096: added the testid for it.

swatigoel (approved):

PR #976, opened 2023-10-31
Feat(Inji-502): Enable user to search issuer from the list in Add New card

Anil-kumar-Majji:
**[INJI-502](https://mosip.atlassian.net/browse/INJI-502)** - As INJI user,  to download the VC, they should be easily able to search the issuer from the list.

[INJI-502]: https://mosip.atlassian.net/browse/INJI-502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


screens/Issuers/IssuersScreen.tsx
```diff
@@ -124,24 +129,56 @@ export const IssuersScreen: React.FC<
     );
   }
 
+  const showAllIssuers = () => {
```
vijay151096, line 132: No need for separate method showallIssuers. we can handle it in the search issuers itself by passing empty text.

screens/Issuers/IssuersScreen.tsx
```diff
@@ -124,24 +129,56 @@ export const IssuersScreen: React.FC<
...
+    setFilteredSearchData(issuers);
+  };
+
+  const searchIssuer = text => {
```
vijay151096, line 136: instead of searchIssuer, can we rename it to filterIssuers

screens/Issuers/IssuersScreen.tsx
```diff
@@ -124,24 +129,56 @@ export const IssuersScreen: React.FC<
...
+    if (text === '') {
```
vijay151096, line 137: Instead of if-else can we add reframe it this way.

let filterdData = issuers;
if (text.length > 1) {
      filterdData = issuers.filter(item => {
        if (
          getDisplayObjectForCurrentLanguage(item.display)
            ?.name.toLowerCase()
            .includes(text.toLowerCase())
        ) {
          return getDisplayObjectForCurrentLanguage(item.display);
        }
      }, []);
    }
setFilteredSearchData(filterdData);
setSearch(text);

screens/Issuers/IssuersScreen.tsx
```diff
@@ -124,24 +129,56 @@ export const IssuersScreen: React.FC<
...
-              marginVertical: 14,
-              marginHorizontal: 9,
-            }}>
-            {t('header')}
-          </Text>
-          <View style={Theme.issuersScreenStyles.issuersContainer}>
+        <Column style={Theme.IssuersScreenStyles.issuerListOuterContainer}>
+          <Row margin="3">
+            <Icon
+              testID="search"
```
vijay151096, line 162: can we use unique name for testId, like searchIssuer. 
because in future we might introduce search for vc's in homescereen as well.

screens/Issuers/IssuersScreen.tsx
```diff
@@ -124,24 +129,56 @@ export const IssuersScreen: React.FC<
...
+
+      setFilteredSearchData(filterdData);
+      setSearch(text);
+    }
+  };
+
   return (
     <React.Fragment>
       {controller.issuers.length > 0 && (
-        <Column style={Theme.issuersScreenStyles.issuerListOuterContainer}>
```
vijay151096, line 130: why are we removing this subHeader from the screen, is it not needed ?

screens/Issuers/IssuersScreen.tsx
Anil-kumar-Majji: I tried, but it wouldn't work to show the list of items when entered into the "**Add new card**" screen.
So followed with onLayout props in TextInput with the above function (**showAllIssuers**)

screens/Issuers/IssuersScreen.tsx
Anil-kumar-Majji: Currently, there is no wireframe for this so I removed based on the text inside of that subheader.
I think it was diff. while implementing a search bar. If required, I will revert it.

locales/ara.json
```diff
@@ -159,7 +159,8 @@
   },
   "IssuersScreen": {
     "title": "أضف بطاقة جديدة",
-    "header": "يرجى اختيار جهة الإصدار المفضلة لديك من الخيارات أدناه لإضافة بطاقة جديدة.",
+    "addCardDescription": "يرجى اختيار جهة الإصدار المفضلة لديك من الخيارات أدناه لإضافة بطاقة جديدة.",
```
vijay151096, line 162: please rename this key from addCardDescription to **SubHeader or description.** in all locales.

screens/Issuers/IssuersScreen.tsx
```diff
@@ -124,10 +129,27 @@ export const IssuersScreen: React.FC<
     );
   }
 
+  const filterIssuers = text => {
+    let filterdData = issuers;
+    if (text.length > 0) {
```
vijay151096, line 134: no need text.length here as we are not passing any arguments for default case where all issuers are listed.
we can use **if(text){} instead of if(text.length > 0)**

screens/Issuers/IssuersScreen.tsx
```diff
@@ -124,10 +129,27 @@ export const IssuersScreen: React.FC<
     );
   }
 
+  const filterIssuers = text => {
```
vijay151096, line 132: rename text to **searchText**

locales/ara.json
Anil-kumar-Majji: If I tries to change text key as "subHeader" also need to change its testID into same name. @swatigoel mentioned that don't change testID. So I remained it as same.

screens/Issuers/IssuersScreen.tsx
Anil-kumar-Majji: Done

screens/Issuers/IssuersScreen.tsx
Anil-kumar-Majji: If I follow this I need to pass an empty string into the respective auto-calling method(i.e., filterIssuers) through the onLayout prop of textInput component.

vijay151096 (approved):
looks good to me.

PR #981, opened 2023-11-01
Refactor[INJI-467] : encrypt the missed data which is stored unencrypted.

vijay151096:


components/LanguageSelector.tsx
```diff
@@ -1,50 +1,55 @@
 import React from 'react';
-import {SUPPORTED_LANGUAGES} from '../i18n';
+import i18n, {SUPPORTED_LANGUAGES} from '../i18n';
```
swatigoel, line 2: There are two instances is `i18n` now. One from import and another one at line no 12 from useTranslation()

vijay151096: removed that additonal i18n instance

swatigoel (approved):

PR #985, opened 2023-11-03
fix(INJI-536):check status failed while downloading vc

srikanth716:
Screenshot_1698994248

### DIFF
--- a/machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
+++ b/machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
@@ -62,7 +62,6 @@ const model = createModel(
     transactionId: '',
     bindingTransactionId: '',
     revoked: false,
-    downloadCounter: 0,
     maxDownloadCount: 10,
     downloadInterval: 5000,
     walletBindingResponse: null as WalletBindingResponse,
@@ -83,6 +82,7 @@ const model = createModel(
       STORE_RESPONSE: (response: VC) => ({response}),
       POLL: () => ({}),
       DOWNLOAD_READY: () => ({}),
+      FAILED: () => ({}),
       GET_VC_RESPONSE: (vc: VC) => ({vc}),
       VERIFY: () => ({}),
       LOCK_VC: () => ({}),
@@ -183,7 +183,6 @@ export const ExistingMosipVCItemMachine =
                     log((_, event) => (event.data as Error).message),
                     'sendDownloadLimitExpire',
                   ],
-                  target: 'checkingStatus',
                 },
               },
             },
@@ -200,6 +199,12 @@ export const ExistingMosipVCItemMachine =
                 DOWNLOAD_READY: {
                   target: 'downloadingCredential',
                 },
+                FAILED: [
+                  {
+                    actions: ['incrementDownloadCounter', 'sendVcUpdated'],
+                    target: 'verifyingDownloadLimitExpiry',
+                  },
+                ],
               },
             },
             downloadingCredential: {
@@ -214,7 +219,12 @@ export const ExistingMosipVCItemMachine =
                     actions: [
                       send('POLL_DOWNLOAD', {to: 'downloadCredential'}),
                       'incrementDownloadCounter',
+                      'sendVcUpdated',
                     ],
+                  },
+                  {
+                    target: 'verifyingDownloadLimitExpiry',
+                    actions: ['sendVcUpdated'],
                   },
                 ],
                 CREDENTIAL_DOWNLOADED: {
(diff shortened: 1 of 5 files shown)


machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
```diff
@@ -1066,8 +1078,19 @@ export const ExistingMosipVCItemMachine =
...
+          vcMetadata: context =>
+            new VCMetadata({
+              ...context.vcMetadata,
+              downloadCounter: context.vcMetadata.downloadCounter + 1,
+            }),
+        }),
+
+        deleteCounterFromVcMetadata: assign({
+          vcMetadata: context => {
+            delete context.vcMetadata.downloadCounter;
```
swatigoel, line 1091: We have to delete downloadCounter or set it 0?

srikanth716: once the VC is completely downloaded we don't need the downloadCounter and also to keep vcMetadata as it is like previously downloadCounter is deleted.

machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
```diff
@@ -1381,6 +1404,11 @@ export const ExistingMosipVCItemMachine =
                 case 'printing':
                   callback(model.events.DOWNLOAD_READY());
                   break;
+                case 'FAILED':
```
vijay151096, line 1407: Since FAILED and default case are having same implementation 

> case 'FAILED':
>default:
>    callback(model.events.FAILED()); 

we can mention it this way.

machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
```diff
@@ -62,7 +62,6 @@ const model = createModel(
     transactionId: '',
     bindingTransactionId: '',
     revoked: false,
-    downloadCounter: 0,
```
vijay151096, line 65: Since, downloadCounter is not common for all VC's and also it is not useful post download (success or failure), so can we mention it as context only and not as field in vcMetadata.

Moreover vcMetadata can be light weight with only the meta info that is common for all VC's should be avl there.

machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
```diff
@@ -214,14 +219,21 @@ export const ExistingMosipVCItemMachine =
...
+                    target: 'verifyingDownloadLimitExpiry',
+                    actions: ['sendVcUpdated'],
+                  },
                 ],
                 CREDENTIAL_DOWNLOADED: {
                   actions: ['setStoreVerifiableCredential', 'storeContext'],
                 },
                 STORE_RESPONSE: {
                   actions: [
+                    'deleteCounterFromVcMetadata',
```
vijay151096, line 235: deleting and adding the vcMetadata creates more unstable values in vcMetadata

machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
```diff
@@ -62,7 +62,6 @@ const model = createModel(
     transactionId: '',
     bindingTransactionId: '',
     revoked: false,
-    downloadCounter: 0,
     maxDownloadCount: 10,
```
vijay151096, line 65: Maximum download count should not be hardcoded, it should be fetched from inji-default.properties.
we can use getAllConfiguration method to get the property and use it here.

shared/VCMetadata.ts
```diff
@@ -13,6 +13,7 @@ export class VCMetadata {
   issuer?: string = '';
   protocol?: string = '';
   static vcKeyRegExp = new RegExp(VC_ITEM_STORE_KEY_REGEX);
+  downloadCounter?: number = 0;
```
vijay151096, line 16: can we remove downloadCounter from vcmetadata ?

machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
srikanth716: maxDownloadCount is assigning from the inji-default.properties

machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
srikanth716: downloadCounter moved to context from vcMetadata

shared/VCMetadata.ts
srikanth716: moved to context

machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
srikanth716: downloadCounter moved to context

vijay151096 (approved):
looks good to me.

swatigoel (approved):

PR #996, opened 2023-11-08
 feat(inji-411): Remove caching for vc files in release

tilak-puli:


swatigoel:
Closing as we don't need it in release-0.10.0

PR #1003, opened 2023-11-13
fix(INJI-552): handle biometric reject on vc download

KiruthikaJeyashankar:
Show error when user cancel biometric during download via issuer

**actions** in error popup after biometric cancellation

- Try again -> ask for biometric and resume downloading from there itself if biometric given
- Cancel -> Go back to add new card screen

**Content**
Do you want to cancel download?
Biometric confirmation is required to continue downloading the card.
[Cancel] [Try again]

**Recording / Screenshots**
image


vijay151096 (approved):
PR looks good.

PR #1004, opened 2023-11-14
fix(inji-522): enable the location service before initiating the vc sharing.

vijay151096:


machines/bleShare/scan/scanMachine.ts
```diff
@@ -679,8 +680,21 @@ export const scanMachine =
           },
         },
         checkingLocationService: {
-          initial: 'checkingPermissionStatus',
+          initial: 'checkLocationService',
```
KiruthikaJeyashankar, line 683: Can we rename the parent state name / child state name accordingly as we have similar name for both states

vijay151096: parent state is named as check**ing**LocationService as it is continuing operation, we named the child as checkLocationService, as it is a operation that is performed currently.

which gives more clarity, pls let me know if you have any other concerns.

android/app/build.gradle
```diff
@@ -104,7 +102,7 @@ android {
 
     namespace 'io.mosip.residentapp'
 
-    
+
     namespace 'io.mosip.residentapp'
```
swatigoel, line 106: Delete this duplicate entry for namespace

swatigoel (changes requested):

android/app/build.gradle
vijay151096: removed.

machines/bleShare/scan/scanMachine.ts
KiruthikaJeyashankar: The state checkLocationService checks if location is turned on or not, where as checkingLocationService is combination of states for check if location turned on -> if yes check if app has permission to access location -> if yes go ahead or request.
Considering this , the parent is performing a more functionality than just checking for location service. Should be bring in more generic state name for the parent considering its functionalities? (like checkLocationStatus)

swatigoel (approved):

machines/bleShare/scan/scanMachine.ts
vijay151096: updated it @KiruthikaJeyashankar

KiruthikaJeyashankar (approved):

PR #1012, opened 2023-11-16
feat(INJI-560): add error events at multiple places 

Alka1703:


screens/Home/MyVcs/AddVcModalMachine.ts
```diff
@@ -330,11 +330,21 @@ export const AddVcModalMachine =
               'OTP is invalid': 'invalidOtp',
               'OTP has expired': 'expiredOtp',
             };
-            return OTP_ERRORS_MAP[message]
+
+            let otpErrorMessage = OTP_ERRORS_MAP[message]
```
swatigoel, line 334: otpErrorMessage should be const as it's not changing.

swatigoel (changes requested):

swatigoel (approved):

PR #1040, opened 2023-11-30
feat(INJI-503): add configurable issuer details

KiruthikaJeyashankar:


components/openId4VCI/Issuer.tsx
```diff
@@ -1,15 +1,13 @@
...
 import testIDProps from '../../shared/commonUtil';
 import {Text} from '../ui';
+import {displayType} from '../../machines/issuersMachine';
 
 export const Issuer: React.FC<IssuerProps> = (props: IssuerProps) => {
-  const {t} = useTranslation('IssuersScreen');
-
   function getIssuerLogo() {
-    return {uri: props.logoUrl};
+    return {uri: props.displayDetails.logo.url};
```
swatigoel, line 10: should we take care of optional in case it happens in future when we fetch data direct from well-known env not from controlled env.
prop. displayDetails?.logo?.url

KiruthikaJeyashankar: Can we make logo as optional when we bring in well-know change? 
Because As of now , this type matches our current structure + when we bring in well-know change we should be changing some props to say if wellKnownLogo available take it or else do something. so can we keep it this way for now @swatigoel ?

swatigoel (approved):

PR #1050, opened 2023-12-01
[Inji 388] : Error message in QR login

Alka1703:


shared/telemetry/TelemetryConstants.js
```diff
@@ -30,6 +30,7 @@ export const TelemetryConstants = {
     hardwareKeyStore:
       'Some security features will be unavailable as hardware key store is not available',
     activationCancelled: 'Activation Cancelled',
+    vcsAreTampered: 'VC are tampered',
```
swatigoel, line 33: Why this change in this commit?

Alka1703: Seems like this was a miss from some previous story and since it was an error event it went highly undetected.

swatigoel (approved):

PR #1068, opened 2023-12-06
ci(INJI-564): use BSD sed to replace strings in pbxproj

vharsh:
- due to some limitations & time constraints I've made changes to *.pbxproj file via `BSD sed` instead of spending more time asking question in forums or in Fastlane GitHub issues

### DIFF
--- a/ios/Inji.xcodeproj/project.pbxproj
+++ b/ios/Inji.xcodeproj/project.pbxproj
@@ -368,10 +368,10 @@
 					"-lc++",
 				);
 				OTHER_SWIFT_FLAGS = "$(inherited) -D EXPO_CONFIGURATION_DEBUG";
-				PRODUCT_BUNDLE_IDENTIFIER = io.mosip.inji.collab;
+				PRODUCT_BUNDLE_IDENTIFIER = io.mosip.inji.mobileid;
 				PRODUCT_NAME = Inji;
 				PROVISIONING_PROFILE_SPECIFIER = "";
-				"PROVISIONING_PROFILE_SPECIFIER[sdk=iphoneos*]" = "match Development io.mosip.inji.collab";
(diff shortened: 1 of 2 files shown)


ios/Inji.xcodeproj/project.pbxproj
```diff
@@ -368,10 +368,10 @@
 					"-lc++",
 				);
 				OTHER_SWIFT_FLAGS = "$(inherited) -D EXPO_CONFIGURATION_DEBUG";
-				PRODUCT_BUNDLE_IDENTIFIER = io.mosip.inji.collab;
+				PRODUCT_BUNDLE_IDENTIFIER = io.mosip.inji.mobileid;
```
vharsh, line 371: this change is required for the sed script to work in a sane manner as it's replacing all instances of `mobileid` with the flavor's correct suffix

vharsh: Maybe a comment above this would be good to state it's importance for the Fastlane & CI.

ios/fastlane/Fastfile
```diff
@@ -115,6 +124,20 @@ platform :ios do
...
+
+    # replace Product name
+    p = "s|PRODUCT_NAME = Inji|PRODUCT_NAME = \"#{generate_app_name}\"|"
+    sh ("sed -i '' -e #{p} ios/Inji.xcodeproj/project.pbxproj")
+
+    # update app display name
+    update_info_plist(
+      plist_path: 'Inji/Info.plist',
+      display_name: "#{generate_app_name}"
+    )
```
vharsh, line 139: these might be accomplished via https://docs.fastlane.tools/actions/update_project_provisioning/ and can be taken for later refactoring IMO https://github.com/fastlane/fastlane/issues/8973 as we are short on time and blocking an internal release

adityankannan-tw (approved):
lgtm

PR #1093, opened 2023-12-13
Inji 569 using svg instead png

srikanth716:


assets/digital_identity_icon.svg
vijay151096, line 1: The Same Naming convention can be followed in all places.
can we use like "Digital_Identity_Icon.svg"

components/ui/themes/PurpleTheme.ts
```diff
@@ -1317,35 +1307,13 @@ export const PurpleTheme = {
   ICON_SMALL_SIZE: 16,
   ICON_MID_SIZE: 22,
   ICON_LARGE_SIZE: 33,
-  PinIcon: require('../../../assets/pin_icon.png'),
   CloseCard: require('../../../assets/card_bg.png'),
-  CardBackground: require('../../../assets/card_bg.png'),
   OpenCard: require('../../../assets/card_bg.png'),
   cardFaceIcon: require('../../../purpleAssets/profile_icon.png'),
```
vijay151096, line 1312: Still Some Images are in png, any specific reason to have this. if not we can still proceed to have svg for this also.

srikanth716: As of now there is no way of using the svg image as an background image, since we are using background images for the closed card and open card preview we still require some png images

swatigoel (changes requested):

components/EditableListItem.tsx
```diff
@@ -32,7 +32,6 @@ export const EditableListItem: React.FC<EditableListItemProps> = props => {
     <ListItem bottomDivider topDivider onPress={() => setIsEditing(true)}>
       <Icon
         name={props.Icon}
-        containerStyle={Theme.Styles.settingsIconBg}
```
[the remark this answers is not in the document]
srikanth716: Icons and Text is not aligned properly due to this styling in settings screen, so removed to make them align properly

assets/Settings_Icon.svg
vijay151096, line 1: can we please check why the settings icon is having the orange theme as default, and upon checkin the branch, i couldn't see any usage of it.

srikanth716: file removed

vijay151096 (approved):
lgtm

swatigoel (approved):

PR #1101, opened 2023-12-18
feat(INJI-622): add telemetry event to track face model init fail/sucess

vharsh:
# Summary

(Q) What we know so far?
(A) The crash is possible on Android & iOS if the Face SDK is not initialized, there could be other reasons as well hence we are going to log an event in Telemetry and observe if a crash still exists.

(Q) Have we verified that the app crashes?
(A) Yes, I have disabled the download of face model and tried to do a Share with Selfie and a Online QR Login on Android & iOS and in both platforms the phones have crashed.

(Q) Could there be other reasons for the crash?
(A) Yes, I’ve observed one of the issue and tried to fix it and I’ll be adding telemetry events to see if the crash still happens when the face SDK model is initialized correctly.

### DIFF
--- a/shared/commonprops/commonProps.ts
+++ b/shared/commonprops/commonProps.ts
@@ -2,6 +2,13 @@ import {configure} from '@iriscan/biometric-sdk-react-native';
 import {changeCrendetialRegistry} from '../constants';
 import {CACHED_API} from '../api';
 import {faceMatchConfig} from '../commonUtil';
+import {
+  getErrorEventData,
+  getImpressionEventData,
+  sendErrorEvent,
+  sendImpressionEvent,
+} from '../telemetry/TelemetryUtils';
+import {TelemetryConstants} from '../telemetry/TelemetryConstants';
 
 export const COMMON_PROPS_KEY: string =
   'CommonPropsKey-' + '6964d04a-9268-11ed-a1eb-0242ac120002';
@@ -16,6 +23,7 @@ export default async function getAllConfigurations(
 
 export async function downloadModel() {
   try {
+    console.log('restart Face model init');
     var injiProp = await getAllConfigurations();
     const maxRetryStr = injiProp.modelDownloadMaxRetry;
     const maxRetry = parseInt(maxRetryStr);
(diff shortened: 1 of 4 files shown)


shared/commonprops/commonProps.ts
```diff
@@ -21,11 +29,29 @@ export async function downloadModel() {
...
+            TelemetryConstants.EndEventStatus.failure,
+          ),
+        );
       }
     }
   } catch (error) {
+    sendErrorEvent(
+      getErrorEventData(
+        TelemetryConstants.FlowType.faceModelInit,
+        TelemetryConstants.EndEventStatus.failure,
```
Alka1703, line 52: This should be an error Id coming from Telemetry constants

shared/commonprops/commonProps.ts
```diff
@@ -21,11 +29,29 @@ export async function downloadModel() {
...
+              TelemetryConstants.FlowType.faceModelInit,
+              TelemetryConstants.EndEventStatus.success,
+            ),
+          );
           break;
         }
+        sendErrorEvent(
+          getErrorEventData(
+            TelemetryConstants.FlowType.faceModelInit,
+            TelemetryConstants.EndEventStatus.failure,
```
Alka1703, line 43: This should the landing screen's name instead of the status. If you want to pass the status send it in additional parameters

shared/commonprops/commonProps.ts
```diff
@@ -21,11 +29,29 @@ export async function downloadModel() {
         var result = await init(resp + '/model.tflite', false);
```
Alka1703, line 29: This is inside a for loop which means we will be logging telemetry events multiple times for the same thing.

Alka1703 (changes requested):

shared/commonprops/commonProps.ts
swatigoel: We should log once after retries count is exhausted

Alka1703 (approved):
LGTM

shared/commonprops/commonProps.ts
```diff
@@ -16,6 +23,7 @@ export default async function getAllConfigurations(
 
 export async function downloadModel() {
   try {
+    console.log('restart Face model init');
```
swatigoel, line 26: Why the log message is `restart`?

vharsh: Should have been removed IMO.

PR #1105, opened 2023-12-19
[Inji 648] add accessibility id to elements in different screens

PuBHARGAVI:


shared/commonUtil.ts
```diff
@@ -20,8 +20,8 @@ export interface Argon2iConfig {
   mode: string;
 }
 
-export default function testIDProps(id) {
-  return isAndroid()
+export default function testIDProps(id, elementType = '') {
+  return isAndroid() || elementType !== ''
```
swatigoel, line 24: Why we need this check here?
For Android, `accessible: true, accessibilityLabel: id` assign id to the component and reflects in appium

swatigoel: If I look at complete code, then looks like it's done for Pressable component only. For Pressable, we have another solution.
Add `accessible={false}` to Pressable component and `accessible={true}` to child component.

PuBHARGAVI: yeah changed this. Yesterday when i tried it didn't work so added it now it is working. 
weird🤣

swatigoel (approved):

PR #1125, opened 2023-12-27
[INJI-620] fix detail view alignment

srikanth716:
eng-ali

### DIFF
--- a/screens/Home/ViewVcModal.tsx
+++ b/screens/Home/ViewVcModal.tsx
@@ -18,19 +18,14 @@ import {
   sendErrorEvent,
 } from '../../shared/telemetry/TelemetryUtils';
 import {TelemetryConstants} from '../../shared/telemetry/TelemetryConstants';
+import {HelpScreen} from '../../components/HelpScreen';
+import {Icon} from 'react-native-elements';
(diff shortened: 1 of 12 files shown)


screens/Home/ViewVcModal.tsx
```diff
@@ -60,7 +55,23 @@ export const ViewVcModal: React.FC<ViewVcModalProps> = props => {
       onDismiss={props.onDismiss}
       headerTitle={t('title')}
       testID="idDetailsHeader"
-      headerElevation={2}>
+      headerElevation={2}
+      arrowLeft={<Icon name={''} />}
```
vijay151096, line 59: Instead of passing empty icon, can't we not pass the arrowLeft param and handle it accordingly.

vijay151096 (approved):
looks good to me.

screens/Home/ViewVcModal.tsx
Anil-kumar-Majji: @vijay151096 This `arrowLeft` prop change will be from my PR. It's the same prop. related changes in Modal.tsx were added from @srikanth716 PR (means this PR).

PR #1126, opened 2023-12-27
[INJI-593]: Text search field as per wire frame in Issuer Screen.

Anil-kumar-Majji:
[**INJI-593**](https://mosip.atlassian.net/browse/INJI-593):  Text search field as per wire frame in Issuer Screen.

- The search bar bottom line will get highlighted when the user focuses on text fields.
-  The cross icon will appear only when the user starts typing and it should clear the searched text by pressing it.


assets/Issuer_search_clearing_button.svg
vijay151096, line 1: Instead of using svg / png images, can we use react native vector icons, which we already use in multiple places.
which will reduce the app bundle size.

**ref:** 
https://oblador.github.io/react-native-vector-icons/
"closecircle" - resembles the image in the wireframe

Anil-kumar-Majji: It worked @vijay151096. Replacement was done

swatigoel (approved):

PR #1132, opened 2024-01-02
[Inji 597] use npm sonar-scanner package to perform sonar qube analysis on inji code

PuBHARGAVI:
- gradlew sonar analysis file of kattu which we were using is performing the sonar qube analysis on the android native code but we wanted it to be applied on inji typescript files so using npm-sonar-analysis.yml file of kattu which is using sonar-scanner npm package


vijay151096 (approved):
lgtm

PR #1139, opened 2024-01-04
Inji 453 restricted uin error message misaligned

srikanth716:


vijay151096 (approved):
PR looks good to me, Since there is  a change in test id, please convey the change to QA automation team, so that automation is not broken.

swatigoel (approved):

PR #1147, opened 2024-01-08
[INJI-692&694] fix overlay height and background border color in purple

srikanth716:
Screenshot_1704714131
Screenshot_1704722982

### DIFF
--- a/components/MessageOverlay.tsx
+++ b/components/MessageOverlay.tsx
@@ -10,11 +10,7 @@ export const MessageOverlay: React.FC<MessageOverlayProps> = props => {
   const {t} = useTranslation('common');
   const style = StyleSheet.create({
     customHeight: {
-      height: props.customHeight
-        ? props.customHeight
-        : props.progress
-        ? 100
-        : 170,
+      minHeight: props.minHeight ? props.minHeight : props.progress ? 100 : 170,
     },
   });
   return (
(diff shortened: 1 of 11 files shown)


components/MessageOverlay.tsx
```diff
@@ -125,5 +121,5 @@ export interface MessageOverlayProps {
   onRetry?: () => void;
   onBackdropPress?: () => void;
   onShow?: () => void;
-  customHeight?: number | string | undefined;
+  minHeight?: number | string | undefined;
```
KiruthikaJeyashankar, line 124: We are passing customHeight prop from other components as well, would this prop name change affect there? Can we rename in consumers as well?

srikanth716: Changed on all the other consumers

components/MessageOverlay.tsx
```diff
@@ -10,11 +10,7 @@ export const MessageOverlay: React.FC<MessageOverlayProps> = props => {
   const {t} = useTranslation('common');
   const style = StyleSheet.create({
     customHeight: {
```
Alka1703, line 12: I still see occurance of customHeight in ScanLayoutController.ts(line 95) and App.tsx(Line 97). Please have a look.

srikanth716: fixed and updated the pr

KiruthikaJeyashankar (approved):

swatigoel (approved):

PR #1161, opened 2024-01-12
[Inji-680]: Changes on VC sharing successful screen added home or history navigation.

Anil-kumar-Majji:
**[INJI-680](https://mosip.atlassian.net/browse/INJI-680)** : [GenderMag]: D22: changes on VC sharing successful screen

[INJI-680]: https://mosip.atlassian.net/browse/INJI-680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


screens/Scan/SuccessfullySharedModal.tsx
```diff
@@ -2,13 +2,14 @@ import React from 'react';
 import {useTranslation} from 'react-i18next';
 import {Theme} from '../../components/ui/styleUtils';
 import {Modal} from '../../components/ui/Modal';
-import {Image} from 'react-native';
-import {Column, Text} from '../../components/ui';
-import {Button} from '../../components/ui';
+import {Pressable} from 'react-native';
```
KiruthikaJeyashankar, line 5: Can we combine these import to avoid duplicate ?
`import {Dimensions, Pressable} from 'react-native';`

screens/Scan/SuccessfullySharedModal.tsx
```diff
@@ -41,13 +42,38 @@ export const SharingSuccessModal: React.FC<
...
+              name="home"
+              color={Theme.Colors.Icon}
+              size={33}
+              containerStyle={
+                Theme.SelectVcOverlayStyles.sharedSuccessfullyIconStyle
+              }
+              onPress={scanLayoutController.DISMISS}
+            />
+            <Text align="center" weight="bold">
+              {t('ScanScreen:status.accepted.home')}
```
KiruthikaJeyashankar, line 59: Can we use useTranslation function (in LINE 17) here rather than sending key ScanScreen?
`t('status.accepted.home')`

screens/Scan/SuccessfullySharedModal.tsx
```diff
@@ -41,13 +42,38 @@ export const SharingSuccessModal: React.FC<
...
+              name="history"
+              color={Theme.Colors.Icon}
+              size={33}
+              containerStyle={
+                Theme.SelectVcOverlayStyles.sharedSuccessfullyIconStyle
+              }
+              onPress={scanLayoutController.GOTO_HISTORY}
+            />
+            <Text align="center" weight="bold">
+              {t('ScanScreen:status.accepted.history')}
```
KiruthikaJeyashankar, line 73: Can we use useTranslation function (in LINE 17) here rather than sending key ScanScreen?
`t('status.accepted.home')`

Also can we change if any similar occurrences are available in this file?

screens/Scan/SuccessfullySharedModal.tsx
```diff
@@ -41,13 +42,38 @@ export const SharingSuccessModal: React.FC<
...
           </Text>
         </Column>
-        <Column margin="0 0  0">
-          <Button
-            type="gradient"
-            title={t('ScanScreen:status.accepted.gotohome')}
-            onPress={scanLayoutController.DISMISS}
-          />
-        </Column>
+        <Row
```
KiruthikaJeyashankar, line 45: Can we also introduce any testIDs for the newly introduced components here?

machines/bleShare/scan/scanMachine.ts
```diff
@@ -441,7 +442,11 @@ export const scanMachine =
                 },
               },
             },
-            navigatingToHistory: {},
+            navigatingToHistory: {
+              invoke: {
+                src: 'disconnect',
```
KiruthikaJeyashankar, line 447: Do we require wallet disconnect on QRLogin as we would not have started the connection with wallet?

screens/Scan/SuccessfullySharedModal.tsx
```diff
@@ -41,13 +42,38 @@ export const SharingSuccessModal: React.FC<
...
+          style={{marginBottom: Dimensions.get('screen').height * 0.06}}>
+          <Pressable>
+            <Icon
+              name="home"
+              color={Theme.Colors.Icon}
+              size={33}
+              containerStyle={
+                Theme.SelectVcOverlayStyles.sharedSuccessfullyIconStyle
+              }
+              onPress={scanLayoutController.DISMISS}
```
KiruthikaJeyashankar, line 56: Can we change this event to `scanLayoutController.DISMISS` -> `scanLayoutController.GO_TO_HOME` as we are navigating to home here

screens/Scan/ScanLayoutController.ts
```diff
@@ -63,6 +63,10 @@ export function useScanLayout() {
   const onStayInProgress = () =>
     scanService.send(ScanEvents.STAY_IN_PROGRESS());
   const onRetry = () => scanService.send(ScanEvents.RETRY());
+  const GOTO_HISTORY = () => {
+    scanService.send(ScanEvents.GOTO_HISTORY());
```
KiruthikaJeyashankar, line 67: Can we change the design here to send DISCONNECT/CLEANUP event which would be listened by machine & that machine invokes wallet disconnect.

This same event can also be sent in GO_TO_HOME, so in the machine we don't need to handle 2 states for wallet disconnection (LINE 617 - 626 in ScanMachine.ts)

screens/Scan/SuccessfullySharedModal.tsx
Anil-kumar-Majji: Done

screens/Scan/SuccessfullySharedModal.tsx
Anil-kumar-Majji: Yes. But in most of components we are use once that useTranslation function.

screens/Scan/SuccessfullySharedModal.tsx
Anil-kumar-Majji: It's better to add everywhere as per the respective flow in all app screens. For any future corrections or re-change of words.

screens/Scan/SuccessfullySharedModal.tsx
Anil-kumar-Majji: Yes, we can. But better to not change right now. Since most of the places in the scanMachine we are using same **DISMISS** event.
For this we may need to make changes in many places which will affect the flow of sharing VC.

screens/Scan/ScanLayoutController.ts
KiruthikaJeyashankar: @vijay151096 we require your inputs here, lets connect on this?

screens/Scan/ScanLayoutController.ts
```diff
@@ -63,6 +63,14 @@ export function useScanLayout() {
   const onStayInProgress = () =>
     scanService.send(ScanEvents.STAY_IN_PROGRESS());
   const onRetry = () => scanService.send(ScanEvents.RETRY());
+  const GOTO_HOME = () => {
```
KiruthikaJeyashankar, line 66: Can we also merge this to something like
```
const GOTO = (screen) => {
    scanService.send(ScanEvents.DISCONNECT());
    navigation.navigate(screen);
  };
```

swatigoel: let's keep it separate for now

machines/bleShare/scan/scanMachine.ts
```diff
@@ -606,7 +606,16 @@ export const scanMachine =
                 },
               },
             },
-            navigatingToHome: {},
+            navigatingToHome: {
```
KiruthikaJeyashankar, line 609: Can we rename this state according to what is performed here?
The state name says that we are navigating to home but we are disconnecting

machines/bleShare/scan/scanMachine.ts
```diff
@@ -595,12 +594,9 @@ export const scanMachine =
             accepted: {
               entry: ['logShared', 'sendVcShareSuccessEvent'],
               on: {
-                DISMISS: {
+                DISCONNECT: {
```
swatigoel, line 597: @Anil-kumar-Majji Don't rename DISMISS by DISCONNECT as DISMISS event is being used from other places.
Apart from this, inner state should be disconnect instead of navigatingToHome as we are not navigating to home

screens/Scan/SuccessfullySharedModal.tsx
Anil-kumar-Majji: Done

KiruthikaJeyashankar (approved):

swatigoel (approved):

PR #1170, opened 2024-01-16
[INJI-699] remove .hmac & .hmace files created due to INJI-612

vharsh:
* also bump up secure-keystore to fix INJI-612

### DIFF
--- a/shared/storage.ts
+++ b/shared/storage.ts
@@ -43,7 +43,7 @@ export const API_CACHED_STORAGE_KEYS = {
   fetchIssuerConfig: (issuerId: string) =>
     `CACHE_FETCH_ISSUER_CONFIG_${issuerId}`,
   fetchIssuerWellknownConfig: (issuerId: string) =>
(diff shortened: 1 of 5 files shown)


shared/storage.ts
```diff
@@ -300,6 +211,13 @@ class Storage {
       const path = getFilePath(key);
       const isFileExists = await FileStorage.exists(path);
       if (isFileExists) {
+        // TEMP(INJI-612-poc): remove .hmac & .hmace files
```
vijay151096, line 214: do we really need these checks, as we have removed the generation of these .hmac and .hmace files, this will end up being a stale code, this code will make sense when inji went live and we want to clean up the user storage to remove the unwanted file, but for now we dont need to do this, as it is intended for testing.

vharsh: makes sense, removing it

vijay151096 (approved):
changes looks good to me.

swatigoel (approved):

PR #1191, opened 2024-01-19
[INJI-714] fix app relaunch in scan & receive screen

KiruthikaJeyashankar:
App relaunchs when nearby permissions is allowed in scan screen and navigated to receive screen and vice versa. This was happening due to our custom logic handled in MainActivity onRequestPermissionsResult performing activity recreation when requestCode is  1 & is not denied. The requestCode for BLUETOOTH_SCAN (in Share screen) permission grant result is 1 (after giving nearby devices permission in receive screen) which satisfied the condition of app relaunch. For this reason onRequestPermissionsResult method has been removed.


swatigoel (approved):

vharsh (approved):
lgtm

PR #1218, opened 2024-01-31
MOSIP-28901 | Added sanity suite for android and ios

kamalsinghthoughtworks:
Added sanity suite for android and ios


vijay151096 (approved):
lgtm

Issue #59, opened 2022-08-03
Resident receiving multiple credential status issuance emails when they download VC using Inji


rakhimosip:
**Describe the bug**
When a resident uses Inji to download a VC they are currently receiving 4 email notifications. Out of that ->
3 emails are sent for credential issuance status
1 email is sent for confirming credential issuance 
Totally the resident receives 4 emails

**To Reproduce**
1. Enroll a resident with an emailID to which you have access and generate UIN
2. Install and setup the Inji app on your smartphone
3. Use the above UIN to download a VC on the app
4. You will receive 4 emails to the configured emailID. Out of that ->
    3 emails are sent for credential issuance status
    1 email is sent for confirming credential issuance 
    The resident receives 4 emails in total.

**Expected behavior**
The resident should receive 2 emails only - 1 email for credential issuance status and 1 email for confirming credential issuance.

Sujithbn:
As of now this issue is not targeted to fix before Beta release.

santhoshsunder:
verified as fixed in the build with commit id in the below screenshot.
![image](https://github.com/mosip/inji/assets/102220709/a33e3a65-94fe-42b3-9735-6c1b36c8fa70)

PR #1228, opened 2024-02-05
INJIMOB-604 VC Verification Failure Scenarios

balachandarg-tw:
- Downloading the VC and showing the status **Valid** only on Verification Success.
- Throwing error for different scenarios for Verification Failure like Technical Error and Network failure error.
- Implemented New Error screen with **Go Back** button which navigates to Home Screen.


machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
```diff
@@ -371,18 +360,50 @@ export const ExistingMosipVCItemMachine =
             src: 'verifyCredential',
             onDone: [
               {
-                actions: ['markVcValid', 'storeContext', 'updateVc'],
-                target: 'idle',
+                actions: [
+                  log((_, event) => 'Verification Success.'),
```
vijay151096, line 364: Can we remove the unwanted loggers

machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
```diff
@@ -371,18 +360,50 @@ export const ExistingMosipVCItemMachine =
...
+                ],
               },
             ],
             onError: [
               {
-                actions: log((_, event) => (event.data as Error).message),
-                target: 'idle',
+                //To-Do Handle Error Scenarios
+                actions: [
+                  log((_, event) => `Verification Error.`),
```
vijay151096, line 375: Can we remove the unwanted loggers

machines/issuersMachine.ts
```diff
@@ -359,15 +359,22 @@ export const IssuersMachine = model.createMachine(
...
               actions: [
-                log((_, event) => (event.data as Error).message),
+                log('Verification Error.'),
+                'resetLoadingReason',
+                'updateVerificationErrorMessage',
                 'sendErrorEndEvent',
               ],
               //TODO: Move to state according to the required flow when verification of VC fails
-              target: 'idle',
+              target: 'handlingCredentialVerificationFailure',
```
vijay151096, line 368: can we rename the state to 'handleCredentialVerificationFailure' instead of 'handlingCredentialVerificationFailure'

screens/Home/MyVcsTab.tsx
```diff
@@ -80,6 +80,13 @@ export const MyVcsTab: React.FC<HomeScreenTabProps> = props => {
   controller.downloadFailedVcs.forEach(vc => {
     failedVCsList.push(`\n${vc.idType}:${vc.id}`);
   });
+
+  const isVerificationError = controller.verificationErrorMessage !== '';
```
vijay151096, line 84: can we rename it as "isVerificationFailed", since we are checking for presence of error, not getting the error.

screens/Home/MyVcsTab.tsx
```diff
@@ -226,16 +233,37 @@ export const MyVcsTab: React.FC<HomeScreenTabProps> = props => {
...
+          testID="verificationError"
+          isVisible={isVerificationError}
+          isModal={true}
+          alignActionsOnEnd
+          title={t('errors.verificationFailed.title')}
+          message={verificationErrorMessage}
+          textButtonEvent={controller.RESET_VERIFY_ERROR}
+          image={SvgImage.ErrorOccurred()}
+          showClose={false}
+          primaryButtonText="goBack"
```
vijay151096, line 247: please use translations for the texts

screens/Home/MyVcsTabController.ts
```diff
@@ -99,5 +111,13 @@ export function useMyVcsTab(props: HomeScreenTabProps) {
 
     REMOVE_TAMPERED_VCS: () => vcService?.send(VcEvents.REMOVE_TAMPERED_VCS()),
     DELETE_VC: () => vcService?.send(VcEvents.DELETE_VC()),
+    RESET_VERIFY_ERROR: () => {
+      console.log('inside reset:::');
```
vijay151096, line 115: can we remove this console.log

shared/vcjs/verifyCredential.ts
```diff
@@ -21,41 +21,81 @@ const ProofPurpose = {
...
+
+    const vcjsOptions = {
+      purpose,
+      suite,
+      credential: verifiableCredential,
+      documentLoader: jsonld.documentLoaders.xhr(),
+    };
+
+    const result = await vcjs.verifyCredential(vcjsOptions);
+    return handleError(result, false);
```
vijay151096, line 60: the methods intention is to handle the verification response, not handling error, can we rename it accordingly.

vijay151096, line 60: do we need to have isException param, as we know if the result is null then it is a exception in all other cases we object result as an object.

machines/issuersMachine.ts
swatigoel: +1 to @vijay151096

machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
```diff
@@ -371,18 +360,50 @@ export const ExistingMosipVCItemMachine =
...
+              ],
+              target: 'idle',
+            },
+            STORE_ERROR: {
+              target: '#vc-item.checkingServerData.savingFailed',
+            },
+          },
+        },
+
+        handlingCredentialVerificationFailure: {
```
swatigoel, line 398: we can have simple name `handleVCVerificationFailure`

machines/issuersMachine.ts
swatigoel: same here rename as `handleVCVerificationFailure`

machines/issuersMachine.ts
```diff
@@ -632,15 +655,20 @@ export const IssuersMachine = model.createMachine(
           0,
         );
       },
-      verifyCredential: async context => {
+      verifyCredential: context => async (callback, onReceive) => {
```
swatigoel, line 658: why we need callback and onReceive?

shared/vcjs/verifyCredential.ts
swatigoel: Rename this method as handleResponse or handleResult

shared/vcjs/verifyCredential.ts
```diff
@@ -21,41 +21,81 @@ const ProofPurpose = {
...
+
+    //ToDo Handle Expiration error message
+  } catch (error) {
+    console.log('Exception during VC Verification->', error);
+    return handleError(null, true);
```
swatigoel, line 65: create error response here and return. No need to send null and isException in handleResponse function

shared/vcjs/verifyCredential.ts
```diff
@@ -21,41 +21,81 @@ const ProofPurpose = {
...
+function handleError(result: any, isException: boolean) {
+  var errorMessage = VerificationErrorType.NO_ERROR;
+  var isVerifiedFlag = true;
 
-  const vcjsOptions = {
-    purpose,
-    suite,
-    credential: verifiableCredential,
-    documentLoader: jsonld.documentLoaders.xhr(),
+  var verificationResult: VerificationResult = {
```
swatigoel, line 73: Construct this as const just before returning as we are changing isVerifiedFlag and errorMessage

machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
```diff
@@ -371,18 +360,50 @@ export const ExistingMosipVCItemMachine =
             src: 'verifyCredential',
             onDone: [
               {
-                actions: ['markVcValid', 'storeContext', 'updateVc'],
-                target: 'idle',
+                actions: [
+                  log((_, event) => 'Verification Success.'),
+                  'setVerifiableCredential',
+                  'markVcValid',
```
swatigoel, line 366: Why do we need markVcValid now as we are not relying on isVerified flag in vc context. We can remove isVerified flag as well

swatigoel (changes requested):

machines/vc.ts
```diff
@@ -214,6 +217,17 @@ export const vcMachine =
             DELETE_VC: {
               target: 'deletingFailedVcs',
             },
+            VERIFY_VC_FAILED: {
+              actions: [
+                'removeVcFromMyVcs',
+                'removeVcFromInProgressDownlods',
+                'setVerificationErrorMessage',
+              ],
```
swatigoel, line 225: At Line #59,  we are accepting only errorMessage with VERIFY_VC_FAILED event but here we are removing VC from myVCs why? as verification as failed it means VC was not there in MyVCs list.
In Existing vc machine, we are sending vcMetadata along with event but not from issuer machine why so? We need to have consistency for handling the events.

swatigoel (changes requested):

machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
PuBHARGAVI: swati right now we are not using it but as part of INJIMOB-695 first acceptance criteria when verifying credentials if there is an network issue and we are not able to verify the vc then also VC should be shown with verify status as pending so we need to use this markVcValid in that scenarios to check if VC is verified or it is in pending state.

machines/vc.ts
PuBHARGAVI: Previously when writing the code we are showing the error message overlay instead of screen and at that time when we click on ok button in the overlay it will remove the error message overlay but in the background the card is stuck in loading state so we had to remove it from myvcs list. Now we are showing screen and when we click on goBack it is going to homescreen and refreshing the vc's so it is not required now we will remove it

PuBHARGAVI: in existing vc machine if the card is not downloaded after retry limit we are showing download failed error message overlay and removing the Vc from vc.ts myvcs context variable and to remove that we are passing vcMetadata along with the event but in issuer machine we don't have such scenario only if the card is downloaded we will show it and if it is failed for some reason we won't be storing it's metadata and details so no need to pass it.

machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
```diff
@@ -371,18 +360,42 @@ export const ExistingMosipVCItemMachine =
...
-                actions: ['markVcValid', 'storeContext', 'updateVc'],
-                target: 'idle',
+                actions: ['setVerifiableCredential', 'storeContext'],
               },
             ],
             onError: [
               {
-                actions: log((_, event) => (event.data as Error).message),
-                target: 'idle',
+                //To-Do Handle Error Scenarios
```
swatigoel, line 368: is To-Do still pending?

shared/vcjs/verifyCredential.ts
```diff
@@ -21,41 +21,81 @@ const ProofPurpose = {
...
+    };
+  }
+}
+
+function handleResponse(result: any) {
+  var errorMessage = VerificationErrorType.NO_ERROR;
+  var isVerifiedFlag = true;
+
+  if (result != null && !result.verified) {
+    if (result['results'][0].error.name == 'jsonld.InvalidUrl') {
```
swatigoel, line 76: we can use if (result?.verified) simplified condition

swatigoel (approved):

PR #1230, opened 2024-02-06
[INJIMOB-799] append timestamp to the restored vc vckey to make it unique

PuBHARGAVI:


shared/storage.ts
```diff
@@ -205,16 +205,37 @@ class Storage {
...
+      let vc = allVCs[key];
+      const unixTimeStamp = Date.now();
+      const prevRequestId = vc.vcMetadata.requestId;
+      vc.vcMetadata.requestId = this.getRestoredVcUpdatedRequestId(
+        unixTimeStamp,
+        vc.vcMetadata.requestId,
+      );
+      dataFromDB.myVCs.map(myVcMetadata => {
+        if (myVcMetadata.requestId === prevRequestId) {
+          myVcMetadata.requestId = vc.vcMetadata.requestId;
```
swatigoel, line 232: We should not change request id for VC. We should only append timestamp to existing key

swatigoel (approved):

shared/storage.ts
```diff
@@ -208,13 +222,25 @@ class Storage {
...
+      let vc = allVCs[key];
+      const currentUnixTimeStamp = Date.now();
+      const prevUnixTimeStamp = vc.vcMetadata.timestamp;
+      vc.vcMetadata.timestamp = currentUnixTimeStamp;
+      dataFromDB.myVCs.forEach(myVcMetadata => {
+        if (
+          myVcMetadata.requestId === vc.vcMetadata.requestId &&
+          myVcMetadata.timestamp === prevUnixTimeStamp
+        ) {
+          myVcMetadata.timestamp = currentUnixTimeStamp;
```
vharsh, line 237: This codechange will not have a side-effect as this is being assigned to the loop variable `myVcMetadata`, 
 so this change will not reflect back to the `allVCs array`. I'll make this change in my PR.

vharsh: I was wrong about the above yesterday as for objects, the above doesn't hold true and it gets modified.

```
> let a = [{c: 1, d: 3, e: "alice"}, {c: 2, d: 4, e: "bob"}]
undefined
> a.forEach(x => x.c+=1)
undefined
> a
[ { c: 2, d: 3, e: 'alice' }, { c: 3, d: 4, e: 'bob' } ]
```
for numbers it's not
```
> let c = [1, 2, 3]
undefined
> c.forEach(x => x+=1)
undefined
> c
[ 1, 2, 3 ]
```

PR #1241, opened 2024-02-10
Remove error toaster when retrying for backup or restore

PoojaBabusing:


components/BackupAndRestoreBannerNotification.tsx
```diff
@@ -54,7 +54,9 @@ export const BackupAndRestoreBannerNotification: React.FC = () => {
         />
       )}
 
-      {backUpController.isBackingUpFailure && backupFailure()}
+      {backUpController.isBackingUpFailure &&
+        !backUpController.isFetchDataFromDB &&
```
KiruthikaJeyashankar, line 58: We are in success / failure state only when backup is success / failure respectively, Why do we require this additional check of if its not fetching from DB?

Once user initiated backup, 
- we move to backingUp state's initial state - backingUp.checkDataAvailability 
- and after backup is success we move to backingUp.success state. 
- If user again initiates backup, we land in backingUp.checkDataAvailability state. So here we are not in backingUp.success state, right? If this is the case why do we need extra check of !backUpController.isFetchDataFromDB? 

Please let me know if I am getting it right here!

components/BackupAndRestoreBannerNotification.tsx
```diff
@@ -65,7 +67,9 @@ export const BackupAndRestoreBannerNotification: React.FC = () => {
           testId={'restoreBackupSuccess'}
         />
       )}
-      {restoreController.isBackUpRestoreFailure && restoreFailure()}
+      {restoreController.isBackUpRestoreFailure &&
+        !restoreController.isDownloadBackupFileFromCloud &&
```
KiruthikaJeyashankar, line 71: Same comment here as well, 
Is the condition - !restoreController.isDownloadBackupFileFromCloud redundant here? as isBackUpRestoreFailure itself would give us failure status?

sree96 (approved):

components/BackupAndRestoreBannerNotification.tsx
PoojaBabusing: resolved

swatigoel (approved):

PR #1247, opened 2024-02-13
Injimob 572 bug received vc deleted

balachandarg-tw:
- When receiving same VC again, due to filter used in moveExistingToTop method, Received Vc list section not showing other VCs, but shows all VCs on reopening the app. As per discussion, we are showing received Vcs at the top of the list, so I removed moveExistingToTop and its condition.
- Added timestamp to recevied VC's metadata to make the VC key unique.


vharsh (approved):
lgtm

PR #1251, opened 2024-02-13
Inji 752 backup to iCloud

sree96:


shared/googleCloudUtils.ts
```diff
@@ -127,45 +154,59 @@ class Cloud {
       };
     }
   }
+  static async downloadUnSyncedBackupFiles(): Promise<Boolean> {
```
vharsh, line 157: Do we have to also sync the injibackup ZIP file across iOS devices? As per the official docs of [downloadFile](https://react-native-cloud-storage.vercel.app/docs/api/CloudStorage#downloadfilepath-scope) the downloadFile function downloads the file in full even if the file has been synced as has the suffix `.icloud`, in this case should we turn off the sync in the backup directory and only let the state machines manage the state of the ZIPs & .injibackup files ?


**Tl; dr** : Should we turn off the iCloud sync for the `inji/` & `mmkv` directory across iOS devices, will they ever conflict with the state of each other devices? Two-way download/syncs can cause issues in the longer run.

**Background** : I'm implementing a feature which looks for `.injibackup` files and then subsequently restores the files on app launch. This might cause some issues if a person uses the wallet from multiple devices and backs up in one and restores in another.

machines/backupAndRestore/backupAndRestoreSetup.ts
```diff
@@ -57,46 +69,109 @@ export const backupAndRestoreSetupMachine = model.createMachine(
...
+              cond: 'isNotSignedInIOSAndViaConfirmationFlow',
+              actions: [
+                () => console.log('isNotSignedInIOSAndViaConfirmationFlow'),
+                'unsetIsLoading',
+                'setErrorReasonAsAccountRequired',
+              ],
+              target: '.error',
+            },
+            {
+              cond: 'isNotSignedInIOS',
```
vharsh, line 143: Can this step ever be reached as one has to setup their iCloud accounts when they start using an iPhone?

shared/googleCloudUtils.ts
sree96: This call is just for the sync only. This method make sures that the .icloud file has been synced properly and on next call to download the .zip file will exist. 

Also yes this method can be used to sync files across the devices with same apple ID logged in. 

Inorder to implement the first launch we still need a sync. Thats why this method was introduced.

machines/backupAndRestore/backupAndRestoreSetup.ts
sree96: There can be scenario when the user logged out after app install or iCloud is disabled for Inji. At any of these instances we will only get `false`. So this condition can be hit multiple scenarios. These are some of the edge cases where this can happen

1. What if I try to setup without an apple account?
2. Deleting / removing the apple account?
4. turnoff iCloud access for Inji
5. turnoff iCloud itself?
6. Logging/signing out of apple account

shared/googleCloudUtils.ts
```diff
@@ -127,45 +154,59 @@ class Cloud {
...
+
     if (!cloudFileName) {
-      const allFiles = await CloudStorage.readdir(
-        `/`,
-        CloudStorageScope.AppData,
-      );
-      cloudFileName = allFiles[0];
+      cloudFileName = (await this.getBackupFilesList()).filter(file =>
+        file.match(this.BACKUP_FILE_REG_EXP),
+      )[0];
```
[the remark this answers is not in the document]
sree96: Yes, It is part of backup upload and is performed after a successful backup upload (https://github.com/mosip/inji/pull/1251/files#diff-0502453ae41068cbe5504359ec5d635f6e7e4a6c647149e62164172c8e14c0f4L206)


This method is just to get the latest uploaded file details from cloud.

machines/backupAndRestore/backupAndRestoreSetup.ts
vharsh: Valid points

swatigoel (approved):

PR #1277, opened 2024-02-20
Inji 604 Verify VC error changes to be merged into release branch

balachandarg-tw:
- Downloading the VC and showing the status Valid only on Verification Success.
- Throwing error for different scenarios for Verification Failure like Technical Error and Network failure error.
- Implemented New Error screen with Go Back button which navigates to Home Screen.


machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
```diff
@@ -413,18 +389,61 @@ export const ExistingMosipVCItemMachine =
...
+          on: {
+            STORE_RESPONSE: {
+              actions: ['sendVerificationError'],
+            },
+          },
         },
+
         checkingVerificationStatus: {
           description:
             'Check if VC verification is still valid. VCs stored on the device must be re-checked once every [N] time has passed.',
```
Alka1703, line 449: Hhat if the cond `isVCValid`(in line 452) is true ? How will we trigger autobackup for that scenario?

Alka1703 (changes requested):

machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
balachandarg-tw: We are not considering isVCValid() method now as its value is always false. We have it to handle other error scenario as part of INJI -695. Refer this comment(https://github.com/mosip/inji/pull/1228#discussion_r1479793926) . So only successful verification , we are storing the VC and trigger auto backup in the current flow.

Alka1703: In case if it  has to return true in the future, please add a TODO for backup handling.

Alka1703 (changes requested):

machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
balachandarg-tw: Yeah sure.

balachandarg-tw: Added

swatigoel (approved):

PR #1283, opened 2024-02-21
[INJIMOB - 865,868] - Fix duplicate VC key for restored VC's and remove metadata if vc file not found

adityankannan-tw:


vharsh (approved):
lgtm

PR #1293, opened 2024-02-26
[INJIMOB -784] - Add face verification consent screen

adityankannan-tw:


machines/QrLoginMachine.ts
```diff
@@ -483,6 +523,11 @@ export const qrLoginMachine =
         },
       },
       guards: {
+        isConsentGiven: context => {
+          console.log('isConsentGiven value', context.showFaceAuthConsent);
```
swatigoel, line 527: Remove this console log

machines/QrLoginMachine.ts
```diff
@@ -483,6 +523,11 @@ export const qrLoginMachine =
...
+          return !context.showFaceAuthConsent;
```
swatigoel, line 528: guard says `isConsentGiven` -> so expectation would be, we will get true if consent is given. So don't use negation

machines/bleShare/scan/scanMachine.ts
```diff
@@ -1210,6 +1267,10 @@ export const scanMachine =
       },
 
       guards: {
+        showFaceAuthConsentScreen: context => {
+          return !context.showFaceAuthConsent;
+        },
```
swatigoel, line 1272: same here guard says showFaceAuthConsentScreen -> so expectation would be to return true.

swatigoel (changes requested):

PR #1314, opened 2024-02-29
[Injimob-855] show error screens in both wallet and verifier when shared vc is not stored properly in the verifier

PuBHARGAVI:


screens/Request/ReceiveVcScreen.tsx
```diff
@@ -83,11 +85,15 @@ export const ReceiveVcScreen: React.FC = () => {
         message={t('saving')}
         progress={true}
       />
-      <ErrorMessageOverlay
-        isVisible={controller.IsSavingFailedInIdle}
-        error={storeErrorTranslationPath}
-        translationPath={'ReceiveVcScreen'}
-        onDismiss={controller.DISMISS}
+
```
Alka1703, line 88: @PuBHARGAVI please make changes to show Received Vc only after it is saved successfully. While saving the progess bar can be shown.

Alka1703 (changes requested):

machines/bleShare/scan/scanMachine.ts
```diff
@@ -608,7 +608,7 @@ export const scanMachine =
             },
             rejected: {
               on: {
-                DISMISS: {
+                RETRY: {
```
vharsh, line 611: If this is changed to retry, should the button triggering this event be also changed or was this change intended for the other, i.e. receiveVC machine?

screens/QrLogin/QrConsent.tsx
```diff
@@ -40,7 +40,8 @@ export const QrConsent: React.FC<QrConsentProps> = props => {
             style={Theme.TextStyles.small}
             weight="bold"
             margin={'10 0 0 0'}>
-            {getClientNameForCurrentLanguage(controller.clientName)} {t('access')}
+            {getClientNameForCurrentLanguage(controller.clientName)}{' '}
+            {t('access')}
```
vharsh, line 44: 👍🏼

swatigoel (approved):

machines/bleShare/scan/scanMachine.ts
PuBHARGAVI: When the shared vc is not stored into the storage then we will show the error status screen on both wallet and verifier and in wallet(in scanLayout.tsx) we are calling SharingStatusModal file for all types of errors and we are passing retry and goto home methods to this file and these methods are common for all error types so to support this i have changed dismiss to retry here.

swatigoel (approved):

PR #1317, opened 2024-03-01
[Inji-604]: Addressing the issue of activity log and data back up being called again during app launch and sharing.

balachandarg-tw:
Fix:
- Removing Verification call during Refresh and adding it only after downloading VC.


machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
```diff
@@ -89,7 +89,6 @@ const model = createModel(
       DOWNLOAD_READY: () => ({}),
       FAILED: () => ({}),
       GET_VC_RESPONSE: (vc: VC) => ({vc}),
-      VERIFY: () => ({}),
```
Alka1703, line 92: Why is this VERIFY event from here?

Alka1703 (changes requested):

machines/VCItemMachine/ExistingMosipVCItem/ExistingMosipVCItemMachine.ts
balachandarg-tw: This event is unused and might have been used for any past features before.

Alka1703: Can you remove it from EsignetMosipVCItemMachine as well?

balachandarg-tw: Removed the event.

Alka1703 (approved):
LGTM

swatigoel (approved):

PR #1337, opened 2024-03-18
[INJIMOB-895] add check for welknowConfig to render the vcSkeleton

PoojaBabusing:


components/VC/Views/VCCardView.tsx
```diff
@@ -58,6 +58,7 @@ export const VCCardView: React.FC<
 
   const [fields, setFields] = useState([]);
   const [wellknown, setWellknown] = useState(null);
+  const [isConfigFetched, setIsConfigFetched] = useState(false);
```
vijay151096, line 61: can't we use wellknown state variable itself to check if it is populated or not ? 

if wellknown != null : 
=> then It is fetched from wellknown
 else 
 => it has not yet populated.

swatigoel (approved):

PR #1340, opened 2024-03-19
[INJI-314] [INJI-900] add jest testing config, mocks and sample unit tests for Inji

vharsh:
# High level summary of changes

1. Add mocks of libraries which were preventing a basic snapshot or unit test from running
    - some had to be done manually as the library authors didn't provide mocks
    - libs which provided setupFiles were imported directly as per their GitHub issue comments or docs
2. Setup Jest with babel & react-native presets
    - TODO: For later, we'd need to decide if `jest-expo` preset would be more useful.
3. Setup code coverage calculation, NOTE: At the moment it's disabled, run `npx jest` for running tests & `npx jest --coverage` for running tests with coverage.
4. [kludge] (**really questionable action**) Bump up Expo version to make the iOS app build-release and run.
    - This needs to be re-looked at a later stage as it's best we don't deviate too much from the expo-compatibility matrix with the help of expo community. Running `npx expo-doctor` gives a good idea to see which deviations need to be fixed.
    - The app has been tested for basic app usage sanity w/o downloading UIN/VIN as I wasn't able to download a VC

## Off-topic changes done for app stability
    
- Some development setup issues need to be looked into and fixed to remove iOS dev-setups issues such as [this](https://github.com/oblador/react-native-vector-icons?tab=readme-ov-file#ios-setup)
- 

# More possible improvements or scope for future work

Mentioned here --> https://mosip.atlassian.net/browse/INJIMOB-900


screens/Settings/SettingScreenController.ts
```diff
@@ -28,20 +28,18 @@ import {
...
 import {RequestRouteProps, RootRouteProps} from '../../routes';
 import {REQUEST_ROUTES} from '../../routes/routesConstants';
-import {isIOS} from '../../shared/constants';
 
 export function useSettingsScreen(props: RootRouteProps & RequestRouteProps) {
   const {appService} = useContext(GlobalContext);
-  const authService = appService.children.get('auth');
-  const settingsService = appService.children.get('settings');
+  const authService = appService?.children?.get('auth');
+  const settingsService = appService?.children?.get('settings');
```
KiruthikaJeyashankar, line 37: const settingsService = appService?.children?.get('settings');
  Can we have this as
`  const settingsService = appService?.children?.get('settings') || {}` so that we need not have this in every usage?

Intro.js
```diff
@@ -0,0 +1,36 @@
+import React, {Component} from 'react';
+import {StyleSheet, Text, View} from 'react-native';
+
```
KiruthikaJeyashankar, line 3: What is the use of this Intro component? Are we using it only for snapshot tests?
If yes, should we have it?

jest.config.js
```diff
@@ -1,3 +1,72 @@
...
+  ],
+  setupFiles: [
+    '<rootDir>/__mocks__/svg.mock.js',
+    '<rootDir>/__mocks__/jest-init.js',
+    '<rootDir>/__mocks__/mmkv-db-setup.js',
+    '<rootDir>/__mocks__/react-native.mock.js',
+    '<rootDir>/__mocks__/telemetry-sdk.mock.js',
+    '<rootDir>/__mocks__/expo-constants.mock.js',
+    '<rootDir>/node_modules/react-native-mmkv-storage/jest/mmkvJestSetup.js',
+    '<rootDir>/node_modules/@react-native-community/netinfo/jest/netinfo-mock.js',
```
swatigoel, line 42: here we are referring node_modules in setupFiles then do we need netInfo in mock folder https://github.com/mosip/inji/pull/1340/files#diff-b49705f2e2ef827d02423b5feba730994fa3ba08d7f39ca464b4648c65826d98

jest.config.js
```diff
@@ -1,3 +1,72 @@
...
+  ],
+  coverageDirectory: 'coverage',
+  moduleNameMapper: {
+    // https://stackoverflow.com/a/54513338
+    '\\.(jpg|ico|jpeg|png|gif|eot|otf|webp|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
+      '<rootDir>/__mocks__/fileMock.js',
+    '\\.(svg)$': '<rootDir>/__mocks__/svg.mock.js',
+    '^\\.\\/locales\\/en\\.json$': '<rootDir>/__mocks__/en.mock.json',
+    '^@digitalbazaar/rsa-verification-key-2018$':
+      '<rootDir>/__mocks__/rsa-verification-key-2018.js',
```
swatigoel, line 68: I don't see this file `rsa-verification-key-2018.js`  in mock folder though there is rsa-verification folder

locales/ara.json
```diff
@@ -25,7 +25,7 @@
     "Wallet": "محفظة"
   },
   "FaceScanner": {
-    "imageCaptureGuide": "أمسك الهاتف بثبات، وحافظ على تركيز وجهك في المنتصف وانقر على ‘التقاط'",
+    "imageCaptureGuide": "أمسك الهاتف بثبات وحافظ على تركيز وجهك في المنتصف",
```
swatigoel, line 28: confirm this with latest develop and revert back if it doesn't match

locales/ara.json
```diff
@@ -97,7 +97,7 @@
     "offlineAuthenticationDisabled": "تفعيل لتسجيل الدخول عبر الإنترنت",
     "inProgress": "في تَقَدم",
     "credentialActivated": "مفعل",
-    "profileAuthenticated": "رمز الاستجابة السريعة تسجيل الدخول",
+    "profileAuthenticated": "رمز الاستجابة السريعة تسجيل الدخول"
```
swatigoel, line 100: confirm this with latest develop and revert back if it doesn't match

shared/constants.ts
```diff
@@ -7,8 +7,8 @@ export let MIMOTO_BASE_URL = MIMOTO_HOST;
 export let ESIGNET_BASE_URL = ESIGNET_HOST;
 export let DEBUG_MODE_ENABLED = DEBUG_MODE === 'true';
 
-export const changeCrendetialRegistry = host => (MIMOTO_BASE_URL = host);
-export const changeEsignetUrl = host => (ESIGNET_BASE_URL = host);
+export const changeCrendetialRegistry = (host: any) => (MIMOTO_BASE_URL = host);
+export const changeEsignetUrl = (host: any) => (ESIGNET_BASE_URL = host);
```
swatigoel, line 12: As host will be always be string, so please have specific string type instead of any

shared/commonUtil.ts
```diff
@@ -39,7 +39,7 @@ export interface Argon2iConfig {
   mode: string;
 }
 
-export default function testIDProps(id) {
+export default function testIDProps(id: string | undefined) {
```
swatigoel, line 42: testIdProps will never be undefined. It will be string always, so please remove undefined from type

components/ui/themes/PurpleTheme.ts
```diff
@@ -1581,7 +1577,7 @@ export const PurpleTheme = {
     // https://ethercreative.github.io/react-native-shadow-generator/
 
     if (level === 0) {
-      return null;
+      return {};
```
swatigoel, line 1580: what's the difference null vs empty object

components/ui/themes/DefaultTheme.ts
```diff
@@ -1573,7 +1573,7 @@ export const DefaultTheme = {
     // https://ethercreative.github.io/react-native-shadow-generator/
 
     if (level === 0) {
-      return null;
+      return {};
```
swatigoel, line 1576: what's the difference null vs empty object

components/ui/themes/PurpleTheme.ts
```diff
@@ -1078,6 +1078,9 @@ export const PurpleTheme = {
       borderRadius: 10,
       backgroundColor: Colors.Purple,
     },
+    plain: {
```
swatigoel, line 1081: Why have we added this new style that too only for Purple Theme?

swatigoel (changes requested):

locales/ara.json
vharsh: This was a JSON validation error, this had caused a lot of worry and confusion to us earlier.

components/ui/themes/PurpleTheme.ts
vharsh: It was added very early on and it's not being used, I'm removing it.
cc @srikanth716

components/ui/themes/PurpleTheme.ts
srikanth716: When we are using the elevation style, it is expecting an object not a null value. For the testing the typescript was showing an error for that. So I have changed to {}

components/ui/themes/DefaultTheme.ts
srikanth716: same as above

PR #1348, opened 2024-03-28
[INJIMOB-990] add a job to set the client id and pass to reusable workflow

adityankannan-tw:


.github/workflows/internal-build.yml
```diff
@@ -81,7 +81,21 @@ on:
           - true
 
 jobs:
+  set-client-id:
+    runs-on: ubuntu-latest
+    outputs:
+      CLIENT_NAME: ${{ steps.client-id.outputs.CLIENT_NAME }}
```
swatigoel, line 87: can we name it CLIENT_ID?

swatigoel (changes requested):

swatigoel (approved):

PR #1361, opened 2024-04-03
[INJIMOB-995]: Add testIDs for automation

srikanth716:


components/VCVerification.tsx
```diff
@@ -20,7 +20,7 @@ export const VCVerification: React.FC = ({wellknown, isVerified}: any) => {
         <React.Fragment>
           <VerifiedIcon />
           <Text
-            testID="verificationStatus"
+            testID="statusValue"
```
swatigoel, line 23: let's rename it back to `verificationStatus` and we will ask QA to change this identifier in automation

components/ui/svg.tsx
```diff
@@ -60,7 +60,7 @@ export class SvgImage {
   }
 
   static kebabIcon() {
-    return <KebabIcon />;
+    return <KebabIcon {...testIDProps('KebabIcon')} />;
```
swatigoel, line 63: send testId to this kebabIcon() function as it can be used at many places with different context

swatigoel (changes requested):

swatigoel (approved):

PR #1373, opened 2024-04-15
[INJIMOB-932]  - PixelPass implementation

sree96:


screens/Scan/ScanScreen.tsx
```diff
@@ -15,6 +15,7 @@ import {isIOS} from '../../shared/constants';
 import {BannerNotificationContainer} from '../../components/BannerNotificationContainer';
 import {SharingStatusModal} from './SharingStatusModal';
 import {SvgImage} from '../../components/ui/svg';
+import {assign} from "xstate";
```
Alka1703, line 18: This import is not getting used it seems. If added intentionally we should ideally avoid using xstate in UI components.

sree96: It was added for some other component I guess. Its not used anyways. Removed it. Plz check

components/BannerNotificationContainer.tsx
```diff
@@ -29,6 +34,18 @@ export const BannerNotificationContainer: React.FC = () => {
         </View>
       )}
 
+      {showQuickShareSuccessBanner && (
+          <View style={{marginTop: 10, marginBottom: 10}}>
```
swatigoel, line 38: Can we move this style to style file

components/BannerNotificationContainer.tsx
```diff
@@ -29,6 +34,18 @@ export const BannerNotificationContainer: React.FC = () => {
...
+            <BannerNotification
+                type={BANNER_TYPE_SUCCESS}
+                message={rt('status.accepted.message')}
+                onClosePress={scanScreenController.DISMISS_QUICK_SHARE_BANNER}
+                key={'activatedVcPopup'}
+                testId={'activatedVcPopup'}
+            />
```
swatigoel, line 45: Why key and testId are activatedVcPopup?

components/QrCodeOverlay.tsx
```diff
@@ -7,69 +7,103 @@ import {Theme} from './ui/styleUtils';
...
+    }
+    return qrData;
+  }
 
+  useEffect(() => {
+    (async () => {
+      const qrString = await getQRData();
+      setQrString(qrString);
+    })();
+  }, []);
```
swatigoel, line 43: There is no dependency in useEffect, do we want to reload this every time?
Every change due to useEffect will cause component re-render

components/QrCodeOverlay.tsx
```diff
@@ -7,69 +7,103 @@ import {Theme} from './ui/styleUtils';
...
-            <Icon
-              {...testIDProps('qrCodeCloseIcon')}
-              name="close"
-              onPress={toggleQrOverlay}
-              color={Theme.Colors.Details}
-              size={32}
-            />
-          </Row>
-          <Centered testID="qrCodeDetails" pY={30}>
+    qrString != '' && (
```
swatigoel, line 47: If qrString is empty, are we expecting to show something to user?
Or we expect qrString will never be empty then we don't need this check

swatigoel (changes requested):

components/QrCodeOverlay.tsx
sree96: There was a discussion around this to show some error icon if the QR could not be generated.

sree96: I think we need UI for same.

components/BannerNotificationContainer.tsx
sree96: Updated plz take a look

components/BannerNotificationContainer.tsx
sree96: Updated plz take a look

components/QrCodeOverlay.tsx
sree96: Updated plz take a look

swatigoel (approved):

PR #1380, opened 2024-04-16
[Injimob-1067] fix the search bar issues

PuBHARGAVI:
- Fix the search bar issues
- Fetching MyVCs and ReceiveVCs metadata and data inside VcMetaMachine itself instead of getting it in vcItemMachine


Alka1703 (approved):
Looks good to me

swatigoel (approved):

PR #1392, opened 2024-04-24
[INJIMOB-1247] When filtering the vcs data list filter and remove downloading vc's from this list to avoid the crash 

PuBHARGAVI:


Alka1703 (approved):
Looks good to me.

swatigoel (approved):

PR #1398, opened 2024-04-26
[INJIMOB-1254] fix VC download via sunbird's life insurance VC credentialType

KiruthikaJeyashankar:


machines/Issuers/IssuersService.ts
```diff
@@ -78,8 +78,17 @@ export const IssuersService = () => {
             TelemetryConstants.Screens.webViewPage,
         ),
       );
+      let supportedScopes: [string];
+      if (Object.keys(context.selectedCredentialType).length === 0) {
```
vijay151096, line 82: since we are always calling the credentialTypes endpoint to fetch the credential, do we need a if else, 
We can directly get the scope from supportedScopes = [context.selectedCredentialType['scope']]; 
Correct me, if anything is missing.

KiruthikaJeyashankar: As of now selectedCredentialType is not set up for MOSIP Credential, if we are required to set it up, we might be needed to hardcode the value to get the MOSIPVerifiable credential & not MockVerififableCredential, to avoid that issue, I am designing it like 
- if a issuer does not have multiple credentialTypes , selectedCredentialType is {}
- While invoking authorization, listening to selected Issuer's supportedScopes in case of empty selectedCredentialType

vijay151096: okay fine, but lets keep track of this in the backlog some where.

vijay151096 (approved):
lgtm

swatigoel (approved):

PR #1401, opened 2024-04-29
[INJIMOB-50]: Fix wrong Otp message of VC activation

abhip2565:


machines/VerifiableCredential/VCItemMachine/VCItemMachine.ts
```diff
@@ -302,11 +302,11 @@ export const VCItemMachine = model.createMachine(
               onError: [
                 {
                   actions: [
-                    'setErrorAsWalletBindingError',
+                    'setErrorAsInvalidOtpError',
```
Alka1703, line 305: We have to add condition to check the error message if it is a binding_auth_failed msg only then we will show invalid input error.

Alka1703 (changes requested):

swatigoel:
not needed as of now

PR #1434, opened 2024-05-08
[Injimob 695] For esignet and sunbird Vc's show banner message based on the verification call

PuBHARGAVI:


shared/vcjs/verifyCredential.ts
```diff
@@ -26,10 +27,17 @@ const ProofPurpose = {
 
 export async function verifyCredential(
   verifiableCredential: VerifiableCredential | Credential,
+  vcMetadata: VCMetadata,
 ): Promise<VerificationResult> {
   try {
     let purpose: PublicKeyProofPurpose | AssertionProofPurpose;
-    switch (verifiableCredential.proof.proofPurpose) {
+    const proof = VCMetadata.fromVC(vcMetadata).isFromOpenId4VCI()
```
vijay151096, line 34: can we not add this check "VCMetadata.fromVC(vcMetadata).isFromOpenId4VCI()" here, lets check it before calling and send only the credential, so verifyCredential Method resposibilty is to just verify and not figuring out where the credential would be.

components/VCVerification.tsx
```diff
@@ -21,10 +21,12 @@ export const VCVerification: React.FC = ({wellknown, isVerified}: any) => {
         {statusIcon}
         <Text
           testID="verificationStatus"
-          color={getTextColor(wellknown, Theme.Colors.Details)}
           style={[
             Theme.Styles.detailsValue,
-            {fontFamily: 'Inter_600SemiBold'},
+            {
+              fontFamily: 'Inter_600SemiBold',
```
sree96, line 27: Lets move this to theme file

sree96 (changes requested):
Plz look at the comments

components/VCVerification.tsx
PuBHARGAVI: Done as part of another PR

machines/Issuers/IssuersGuards.ts
```diff
@@ -3,9 +3,12 @@ import {ErrorMessage, Issuers, OIDCErrors} from '../../shared/openId4VCI/Utils';
 import {isHardwareKeystoreExists} from '../../shared/cryptoutil/cryptoUtil';
 import {BiometricCancellationError} from '../../shared/error/BiometricCancellationError';
 import {NETWORK_REQUEST_FAILED, REQUEST_TIMEOUT} from '../../shared/constants';
+import {VerificationErrorType} from '../../shared/vcjs/verifyCredential';
 
 export const IssuersGuards = () => {
   return {
+    isPendingVerificationError: (_context, event) =>
```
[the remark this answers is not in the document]
PuBHARGAVI: Done as part of another pr

shared/vcjs/verifyCredential.ts
PuBHARGAVI: Done as part of another PR

PR #1437, opened 2024-05-09
[INJIMOB-695] for esignet and sunbird show the banner message based on the verification call response

PuBHARGAVI:


machines/Issuers/IssuersMachine.ts
```diff
@@ -326,31 +326,30 @@ export const IssuersMachine = model.createMachine(
...
         },
       },
 
       handleVCVerificationFailure: {
-        on: {
-          RESET_VERIFY_ERROR: {
-            actions: ['resetVerificationErrorMessage'],
-          },
-        },
+        entry: 'sendVerificationError',
```
KiruthikaJeyashankar, line 352: Why are we removing this event?
RESET_VERIFY_ERROR -> this event is required to dismiss the error screen, right?

swatigoel (approved):

PR #1521, opened 2024-06-20
[INJIMOB 797]:  Verify restored VCs

Alka1703:
## Description

> Verify the VCs during backup restoration

>  [INJIMOB-797](https://mosip.atlassian.net/browse/INJIMOB-797)

[INJIMOB-797]: https://mosip.atlassian.net/browse/INJIMOB-797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shared/storage.ts
+++ b/shared/storage.ts
@@ -32,6 +32,9 @@ import {TelemetryConstants} from './telemetry/TelemetryConstants';
 import {BYTES_IN_MEGABYTE} from './commonUtil';
 import fileStorage from './fileStorage';
 import {DocumentDirectoryPath, ReadDirItem} from 'react-native-fs';
+import {verifyCredential} from './vcjs/verifyCredential';
+import {Credential} from '../machines/VerifiableCredential/VCMetaMachine/vc';
+import {isMosipVC} from './Utils';
 
 export const MMKV = new MMKVLoader().initialize();
 const {RNSecureKeystoreModule} = NativeModules;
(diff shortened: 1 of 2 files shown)


shared/storage.ts
```diff
@@ -312,34 +315,62 @@ class Storage {
...
+        allVCKeys.map(async key => {
+          let vc = allVCs[key];
+          const ts = Date.now();
+          const prevUnixTimeStamp = vc.vcMetadata.timestamp;
+
+          const isVerified = await Storage.verifyCredential(
+            vc.verifiableCredential?.credential || vc.verifiableCredential,
+            vc.vcMetadata.issuer,
+          );
+          console.log('isVerified: ', isVerified);
```
swatigoel, line 350: Do we need this logging?

shared/storage.ts
```diff
@@ -312,34 +315,62 @@ class Storage {
...
-          encryptionKey,
-          JSON.stringify(vc),
-        );
-        // Save the VC to disk
-        await this.setItem(updatedVcKey, encryptedVC, encryptionKey);
-      });
+      // 2. Verify the VC and update the result
+
+      await Promise.all(
+        allVCKeys.map(async key => {
```
swatigoel, line 341: Why are we using map here. We are not returning any object. This processing can be achieved by forEach.

shared/storage.ts
```diff
@@ -312,34 +315,62 @@ class Storage {
...
+          let vc = allVCs[key];
```
swatigoel, line 342: We can use const for vc as vc is not changed afterwards.

shared/storage.ts
```diff
@@ -312,34 +315,62 @@ class Storage {
...
-        const encryptedVC = await encryptJson(
-          encryptionKey,
-          JSON.stringify(vc),
-        );
-        // Save the VC to disk
-        await this.setItem(updatedVcKey, encryptedVC, encryptionKey);
-      });
+      // 2. Verify the VC and update the result
+
+      await Promise.all(
```
swatigoel, line 340: Why are we using Promise.all?
Where is this promise resolved?

swatigoel (changes requested):

shared/storage.ts
Alka1703: The difference between `promise.all()` + `array.map()` and `.forEach()` is that it's impossible to await a `forEach()` itself. We can run tasks in parallel in `.forEach()`  just like you can with `.map()`, but we can't wait for all of those parallel tasks to finish, then do something once they've all finished. The whole point of using `.map()` instead of `.forEach()`  is so you can get a list of promises, collect them with `Promise.all()`, then await the whole thing. 

Since we are calling the `verifyCredential()` which is a async method, we have to use `.map()` with `Promise.all()`

shared/storage.ts
Alka1703: Missed this. Removing it.

shared/storage.ts
Alka1703: Overlooked this, as did not change this part. Making the change.

shared/storage.ts
Alka1703: First question is answered in the previous comment. 
Second Question: We just want all the async operations to complete and the VCs and metadata get updated with the verifyCredential value before poceeding further. Promise.all() ensures the completion of all the promises inside it.

swatigoel (approved):

PR #1562, opened 2024-07-15
[Injimob 1540] enhancements for liveness detection in face verification flow

PuBHARGAVI:
## Description

> This PR has the below enhancements for the Liveness Detection in face verification flow:

- Oval shape is removed from the Face scanner in the UI.
- **Max number of times the user can retry the face verification if it is fails** is set to **3** and UI message and translations are updated for the success and failure messages as per the wireframes.
- Added logic to remove used images from the application's cache in order to minimize the amount of space taken up by the images and prevent application crashes.

> [INJIMOB-1540](https://mosip.atlassian.net/browse/INJIMOB-1540)


components/FaceScanner/FaceScanner.tsx
```diff
@@ -100,27 +103,16 @@ export const FaceScanner: React.FC<FaceScannerProps> = props => {
...
-      setInfoText(t('faceProcessingInfo'));
-
-      const result = await cropEyeAreaFromFace(
-        picArray,
-        props.vcImage,
-        faceToCompare,
-      );
-      return result ? props.onValid() : props.onInvalid();
-    } else if (faces.length > 0) {
+    if (counter < MAX_COUNTER) {
```
sree96, line 106: Where we are incrementing the counter?

sree96 (changes requested):
Plz provide inputs for the comment

components/FaceScanner/FaceScanner.tsx
PuBHARGAVI: We are incrementing it in line no 126 @sree96. If the captured frame is satisfying the conditions we will consider it as valid frame and then increment it counter.

we are capturing only 15 frames in liveness detection flow so when we are capturing the 15th frame counter will be 14 and the if condition will be satisfied and if it is valid we will increment the counter to 15 and after capturing 15 it will go into else part and in else part we are freezing the camera so there is no use in incrementing the counter here so i have removed it from else part.

PR #1746, opened 2024-12-17
[INJIMOB-2571]: refactor storage class

Alka1703:


shared/storage.ts
```diff
@@ -573,5 +285,20 @@ export async function isMinimumLimitForBackupRestorationReached() {
   // APIs:
   // 1. CloudStorage.stat(file, context)
   // 2. getUncompressedSize()
-  return await Storage.isMinimumLimitReached('minStorageRequired');
+  return await isMinimumLimitReached('minStorageRequired');
```
KiruthikaJeyashankar, line 288: Storage.isMinimumLimitReached was explicitly stating that we are checking storage related limit
But post moving isMinimumLimitReached function outside Storage class this mapping that its related to storage is missing, can we rename function accordingly to state that this function checks storage limit?

swatigoel: Agree with @KiruthikaJeyashankar

shared/storage.ts
```diff
@@ -52,111 +46,14 @@ async function generateHmac(
...
-        );
-        await this.setItem(key, encryptedValue, encryptionKey);
-        return true;
-      });
-    } catch (error) {
-      console.error('Error while loading backup data ', error);
-      return error;
-    }
+  static backupData = async (encryptionKey: string) => {
+    return await backupData(encryptionKey)
```
swatigoel, line 51: This backupData function here causes confusion. As per import it's exportData but by looking at this line, it looks like recursion here.
so better to rename exportData data function and then use here

shared/storage.ts
```diff
@@ -52,111 +46,14 @@ async function generateHmac(
...
   };
+ 
+  static async restoreBackedUpData(data: string, encryptionKey: string): Promise<any> {
```
swatigoel, line 54: no need to use word backedUp, we can keep it simple restoreBackupData

shared/storage.ts
```diff
@@ -573,5 +285,20 @@ export async function isMinimumLimitForBackupRestorationReached() {
   // APIs:
   // 1. CloudStorage.stat(file, context)
   // 2. getUncompressedSize()
-  return await Storage.isMinimumLimitReached('minStorageRequired');
+  return await isMinimumLimitReached('minStorageRequired');
 }
+
+export async function  isMinimumLimitReached(limitInMB: string) {
```
swatigoel, line 291: rename this function to isMinimumStorageLimitReached?

shared/backupUtils/restoreData.ts
```diff
@@ -0,0 +1,275 @@
...
+      }),
+    );
+  }
+
+  /**
+   * Verify the restored credential
+   * @param vcData  Verifiable Credential from backup
+   * @param vcMetadata  VC metadata
+   */
+  async function verifyCredential(vcData: any, vcMetadata: any) {
```
swatigoel, line 199: verifyCredential can we keep this function in util and use in all flows like download credential, restore

swatigoel (changes requested):

shared/storage.ts
Alka1703: Renamed it to `isMinimumStorageLimitReached`

shared/storage.ts
Alka1703: Missed renaming here. Did it now.

shared/storage.ts
Alka1703: Sure, Updating it.

shared/backupUtils/restoreData.ts
Alka1703: Done

shared/Utils.ts
```diff
@@ -53,4 +56,22 @@ export const parseJSON = (input: any) => {
 
 export const isNetworkError = (error: string) => {
   return error.includes(NETWORK_REQUEST_FAILED);
-};
\ No newline at end of file
+};
+
+export async function verifyCredentialData(
```
swatigoel, line 61: instead of this Utils, move it to Utils under OpenIdVCI.

swatigoel (changes requested):

shared/Utils.ts
Alka1703: Done

swatigoel (approved):

PR #1813, opened 2025-02-24
[INJIMOB-2985] make authorization_servers optional and add fallback logic

abhip2565:
## Description

> make authorization_servers optional and add fallback logic

> [INJIMOB-2985](https://mosip.atlassian.net/browse/INJIMOB-2985&#41)

[INJIMOB-2985]: https://mosip.atlassian.net/browse/INJIMOB-2985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/machines/Issuers/IssuersService.ts
+++ b/machines/Issuers/IssuersService.ts
@@ -59,20 +59,26 @@ export const IssuersService = () => {
     fetchAuthorizationEndpoint: async (context: any) => {
       const wellknownResponse = context.selectedIssuerWellknownResponse;
       const credentialIssuer = wellknownResponse['credential_issuer'];
-      const authorizationServers =
-        wellknownResponse['authorization_servers'] || [credentialIssuer];
-    
+      const authorizationServers = wellknownResponse[
+        'authorization_servers'
+      ] || [credentialIssuer];
+
       const SUPPORTED_GRANT_TYPES = ['authorization_code'];
-    
+      const DEFAULT_AUTHORIZATION_SERVER_SUPPORTED_GRANT_TYPES = [
+        'authorization_code',
+        'implicit',
+      ];
+
       for (const server of authorizationServers) {
         try {
           const authorizationServersMetadata =
             await CACHED_API.fetchIssuerAuthorizationServerMetadata(server);
-    
+
           if (
-            (authorizationServersMetadata?.['grant_types_supported'] || []).some(grant =>
-              SUPPORTED_GRANT_TYPES.includes(grant),
-            )
+            (
+              authorizationServersMetadata?.['grant_types_supported'] ||
+              DEFAULT_AUTHORIZATION_SERVER_SUPPORTED_GRANT_TYPES
+            ).some(grant => SUPPORTED_GRANT_TYPES.includes(grant))
           ) {
             return authorizationServersMetadata['authorization_endpoint'];
           }
@@ -80,9 +86,9 @@ export const IssuersService = () => {
           console.error(`Failed to fetch metadata for ${server}:`, error);
         }
       }
-    
+
       throw new Error(
-        OIDCErrors.AUTHORIZATION_ENDPOINT_DISCOVERY.GRANT_TYPE_NOT_SUPPORTED,
+        OIDCErrors.AUTHORIZATION_ENDPOINT_DISCOVERY.FAILED_TO_FETCH_AUTHORIZATION_ENDPOINT,
       );
     },
 
--- a/machines/Issuers/IssuersService.ts
+++ b/machines/Issuers/IssuersService.ts
@@ -1,8 +1,8 @@
 import NetInfo from '@react-native-community/netinfo';
-import { NativeModules } from 'react-native';
-import { authorize } from 'react-native-app-auth';
+import {NativeModules} from 'react-native';
+import {authorize} from 'react-native-app-auth';
 import Cloud from '../../shared/CloudBackupAndRestoreUtils';
-import { CACHED_API } from '../../shared/api';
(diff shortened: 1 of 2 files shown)


swatigoel (approved):

machines/Issuers/IssuersService.ts
```diff
@@ -61,26 +61,35 @@ export const IssuersService = () => {
        * Incase of multiple entries of authorization_servers, each element is iterated and metadata check is made for support with wallet.
        * For now, its been kept as getting first entry and checking for matching grant_types_supported
```
KiruthikaJeyashankar, line 62: Update the comment as per the logic implemented to select auth_server

KiruthikaJeyashankar (changes requested):

machines/Issuers/IssuersService.ts
abhip2565: removed comment as impl is now as per spec

machines/Issuers/IssuersService.ts
```diff
@@ -57,30 +57,35 @@ export const IssuersService = () => {
...
+    
+          if (
+            (authorizationServersMetadata?.['grant_types_supported'] || []).some(grant =>
+              SUPPORTED_GRANT_TYPES.includes(grant),
+            )
+          ) {
+            return authorizationServersMetadata['authorization_endpoint'];
+          }
+        } catch (error) {
+          console.error(`Failed to fetch metadata for ${server}:`, error);
```
KiruthikaJeyashankar, line 80: Anything reg failure of getting auth server metadata should be considered as GRANT_TYPE_NOT_SUPPORTED, is it?

KiruthikaJeyashankar (changes requested):

machines/Issuers/IssuersService.ts
abhip2565: As per the scope of the function to identify specific errors and handling separately seems overkill to me. Do we want to throw something more generic?
Like `grant types not supported or api request failed`

machines/Issuers/IssuersService.ts
```diff
@@ -57,30 +57,35 @@ export const IssuersService = () => {
...
-          OIDCErrors.AUTHORIZATION_ENDPOINT_DISCOVERY.GRANT_TYPE_NOT_SUPPORTED,
-        );
+    
+      for (const server of authorizationServers) {
+        try {
+          const authorizationServersMetadata =
+            await CACHED_API.fetchIssuerAuthorizationServerMetadata(server);
+    
+          if (
+            (authorizationServersMetadata?.['grant_types_supported'] || []).some(grant =>
```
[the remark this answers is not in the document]
abhip2565: done

machines/Issuers/IssuersService.ts
KiruthikaJeyashankar: In previous flow, we had network error and grant types not supported error separately

But what happens now in case of network failure is L70 makes api call which is wrapped in try block so error comes to catch block, In catch block its just logging of error

Post the catch block in L84 grant types not supported error  is thrown, so will it become like even if network error is obtained, error gets shown up as GRANT_TYPE_NOT_SUPPORTED (which is not true)

abhip2565: As we are iterating through servers, some of them might be non-operational and our network request might fail. In that case we are moving to check the next server, as its the discretion of the wallet to decide its compatibility with auth server.
So, it's difficult to distinguish whats failing.(Network error or server error)

KiruthikaJeyashankar: Got it, so as of now whatever the error while fetching auth server, grant types not supported will be shown up 
We can mention the same in card as well 👍

KiruthikaJeyashankar (approved):

machines/Issuers/IssuersService.ts
```diff
@@ -57,30 +57,35 @@ export const IssuersService = () => {
...
-          OIDCErrors.AUTHORIZATION_ENDPOINT_DISCOVERY.GRANT_TYPE_NOT_SUPPORTED,
-        );
+    
+      for (const server of authorizationServers) {
+        try {
+          const authorizationServersMetadata =
+            await CACHED_API.fetchIssuerAuthorizationServerMetadata(server);
+    
+          if (
+            (authorizationServersMetadata?.['grant_types_supported'] || ["authorization_code", "implicit"]).some(grant =>
```
swatigoel, line 73: extract `["authorization_code", "implicit"]` as constant outside for loop.
It looks complex to understand if condition. If we name it defaultGrantTypeSupported, it will be more readable.

machines/Issuers/IssuersService.ts
```diff
@@ -57,30 +57,35 @@ export const IssuersService = () => {
...
+          }
+        } catch (error) {
+          console.error(`Failed to fetch metadata for ${server}:`, error);
+        }
       }
-
-      return authorizationServerMetadata['authorization_endpoint'];
+    
+      throw new Error(
+        OIDCErrors.AUTHORIZATION_ENDPOINT_DISCOVERY.GRANT_TYPE_NOT_SUPPORTED,
```
swatigoel, line 85: I feel, instead of giving `GRANT_TYPE_NOT_SUPPORTED` we should have generic error like failed to fetch data.
Definite error should be used if we will get that error only.

swatigoel (changes requested):

swatigoel (approved):

PR #1831, opened 2025-02-28
[INJIMOB-2886] add sonar support for inji-wallet

KiruthikaJeyashankar:
Local sonar analysis snapshot

image


sonar-project.properties
KiruthikaJeyashankar, line 1: This file is deleted as sonar-project.properties file will be created in workflow

swatigoel (approved):

PR #1832, opened 2025-03-03
[INJIMOB-3027] add fallback if display property of credential type is missing

abhip2565:
## Description

> if display property of credential type is missing we would fallback to `type` property for `name`.

> Example : [INJIMOB-3027](https://mosip.atlassian.net/browse/INJIMOB-3027&#41)
.

[INJIMOB-3027]: https://mosip.atlassian.net/browse/INJIMOB-3027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/openId4VCI/CredentialType.tsx
+++ b/components/openId4VCI/CredentialType.tsx
@@ -7,11 +7,12 @@ import {displayType} from '../../machines/Issuers/IssuersMachine';
 import {SvgImage} from '../ui/svg';
 import {getDisplayObjectForCurrentLanguage} from '../../shared/openId4VCI/Utils';
 import {CredentialTypes} from '../../machines/VerifiableCredential/VCMetaMachine/vc';
+import { getCredentialType } from '../VC/common/VCUtils';
 
 export const CredentialType: React.FC<CredentialTypeProps> = props => {
   const selectedIssuerDisplayObject = props.item.display?.length
     ? getDisplayObjectForCurrentLanguage(props.item.display)
-    : {};
+    : {name:getCredentialType(props.item)};
 
   return (
     <Pressable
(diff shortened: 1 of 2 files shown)


components/openId4VCI/CredentialType.tsx
```diff
@@ -11,7 +11,7 @@ import {CredentialTypes} from '../../machines/VerifiableCredential/VCMetaMachine
 export const CredentialType: React.FC<CredentialTypeProps> = props => {
   const selectedIssuerDisplayObject = props.item.display?.length
     ? getDisplayObjectForCurrentLanguage(props.item.display)
-    : {};
+    : {name:props.item.credential_definition.type[1]};
```
swatigoel, line 14: instead of picking 1th element, we should filter out default values
VerifiableCredential, VerifiableAttestation and then pick the one from remaining

swatigoel: Same logic we might need to fix here as well https://github.com/mosip/inji-wallet/blob/develop/components/VC/common/VCUtils.tsx#L260-L262

cc: @KiruthikaJeyashankar

swatigoel (changes requested):

components/openId4VCI/CredentialType.tsx
abhip2565: Just thinking what should be our filtering logic? 
According to spec, first element should be VerifiableCredentials. But on the rest there is no specific restrictions.

swatigoel (approved):

PR #1928, opened 2025-05-09
[INJIMOB-3153]: update logic to only send signature for both android and ios for es256 signature creation

Alka1703:


shared/cryptoutil/cryptoUtil.ts
```diff
@@ -299,8 +298,12 @@ export async function createSignatureECR1(
     {format: 'compact', fields: header},
     {key, reference: false},
   );
-  const jws = await signer.update(JSON.stringify(payload)).final();
-  return jws;
+  const jws = await signer.update(JSON.stringify(payload), 'base64').final();
```
abhip2565, line 301: This should be `jwt` imo.

shared/cryptoutil/cryptoUtil.ts
```diff
@@ -299,8 +298,12 @@ export async function createSignatureECR1(
...
+  const parts = jws.split('.');
```
abhip2565, line 302: `jwtParts` could give more context specific clarity i think instead of _parts_.

shared/cryptoutil/cryptoUtil.ts
```diff
@@ -299,8 +298,12 @@ export async function createSignatureECR1(
...
+  if (parts.length !== 3) {
+    throw new Error('Invalid JWS format');
```
abhip2565, line 304: Jwt instead of jws

abhip2565 (changes requested):

swatigoel (approved):

shared/cryptoutil/cryptoUtil.ts
abhip2565: We would be using jws as payload to sign can contain generic claims not specific jwt.

shared/cryptoutil/cryptoUtil.ts
abhip2565: lets make it jwsParts.

abhip2565 (approved):
LGTM

PR #1947, opened 2025-05-29
Injimob 3186 - Added TTL to the cache

Sairam-g9162:


machines/app.ts
```diff
@@ -532,6 +535,18 @@ interface AppInfo {
 
 type State = StateFrom<typeof appMachine>;
 
+const updateCacheTTLFromConfig = async () => {
+  const response = await getAllConfigurations(undefined, false);
+  console.log(response);
```
abhip2565, line 540: Can we remove console logs.

shared/api.ts
```diff
@@ -232,26 +251,33 @@ async function generateCacheAPIFunctionWithCachePreference(
   onErrorHardCodedValue?: any,
 ) {
   try {
-    const response = await getItem(cacheKey, null, '');
+    const cachedData = await getItem(cacheKey, null, '');
 
-    if (response) {
-      return response;
+    if (cachedData && isCacheValid(cachedData)) {
+      console.log('valid cache found ');
```
abhip2565, line 257: Lets remove

shared/InitialConfig.ts
```diff
@@ -18,5 +18,6 @@ export const INITIAL_CONFIG = {
     aboutInjiUrl: 'https://docs.mosip.io/inji/inji-wallet/inji-mobile',
     faceSdkModelUrl: '',
     openId4VCIDownloadVCTimeout: '30000',
+    cacheTTL: '3600000',
```
abhip2565, line 21: can we mention the unit in variable name, to make it more intuitive.

machines/app.ts
swatigoel: +1 to @abhip2565

machines/app.ts
```diff
@@ -532,6 +535,18 @@ interface AppInfo {
...
+  if (response?.cacheTTL && typeof response.cacheTTL === 'number') {
+    changeCacheTTL(response.cacheTTL);
```
swatigoel, line 542: instead of `changeCacheTTL`, let's name function as `updateCacheTTL`

machines/app.ts
```diff
@@ -532,6 +535,18 @@ interface AppInfo {
...
+  } else {
+    console.log(
```
swatigoel, line 544: instead of `console.log` we have `console.error` or `console.warn`, please check those

shared/api.ts
```diff
@@ -232,26 +251,33 @@ async function generateCacheAPIFunctionWithCachePreference(
...
+      return cachedData.response;
     } else {
       const response = await fetchCall();
-      setItem(cacheKey, response, '').then(() =>
-        console.log('Cached response for ' + cacheKey),
+      const cacheObject = createCacheObject(response);
+      setItem(cacheKey, cacheObject, '').then(() =>
+        console.log(
```
swatigoel, line 263: it should be `console.debug` or `console.info`

shared/api.ts
```diff
@@ -266,25 +292,32 @@ async function generateCacheAPIFunctionWithAPIPreference(
 ) {
   try {
     const response = await fetchCall();
-    setItem(cacheKey, response, '').then(() =>
-      console.log('Cached response for ' + cacheKey),
+    const cacheObject = createCacheObject(response);
+    setItem(cacheKey, cacheObject, '').then(() =>
+      console.log(
```
swatigoel, line 297: it should be `console.debug` or `console.info`

shared/constants.ts
```diff
@@ -12,6 +12,7 @@ import {KeyTypes} from './cryptoutil/KeyTypes';
 export let MIMOTO_BASE_URL = MIMOTO_HOST;
 export let ESIGNET_BASE_URL = ESIGNET_HOST;
 export let DEBUG_MODE_ENABLED = DEBUG_MODE === 'true';
+export let DEFAULT_CACHE_TTL = 60 * 60 * 1000;
```
swatigoel, line 15: instead of naming it `DEFAULT_CACHE_TTL`, name it `CACHE_TTL` as we are updating this value based on config.

swatigoel (changes requested):

machines/app.ts
Sairam-g9162: console log has been  removed

shared/api.ts
Sairam-g9162: console log has been removed

machines/app.ts
```diff
@@ -510,6 +512,7 @@ export const appMachine = model.createMachine(
       },
 
       generateKeyPairsAndStoreOrder: async () => {
+        updateCacheTTLFromConfig();
```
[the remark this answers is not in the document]
Sairam-g9162: Here we are doing a function call because as soon as the fresh installation of the app is done this function helps to make the api call to fetch all properties and  update cacheTTL that will be fetched from config

machines/app.ts
Sairam-g9162: function name changed to updateCacheTTL

machines/app.ts
Sairam-g9162: updated with console.warn

shared/api.ts
Sairam-g9162: updated with console.info

shared/api.ts
Sairam-g9162: updated with console.info

shared/api.ts
```diff
@@ -232,26 +251,33 @@ async function generateCacheAPIFunctionWithCachePreference(
...
+        ),
       );
 
       return response;
     }
   } catch (error) {
     console.warn(`Failed to load due to network issue in cache preferred api call.
      cache key:${cacheKey} and has onErrorHardCodedValue:${
-      onErrorHardCodedValue != undefined
+      onErrorHardCodedValue !== undefined
```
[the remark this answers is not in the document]
Sairam-g9162: thanks for the correction, updated to !=

shared/api.ts
```diff
@@ -266,25 +292,32 @@ async function generateCacheAPIFunctionWithAPIPreference(
...
+          ' with expiry: ' +
+          new Date(cacheObject.expiry).toISOString(),
+      ),
     );
     return response;
   } catch (error) {
     console.warn(`Failed to load due to network issue in API preferred api call.
      cache key:${cacheKey} and has onErrorHardCodedValue:${
-      onErrorHardCodedValue != undefined
+      onErrorHardCodedValue !== undefined
```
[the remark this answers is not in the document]
Sairam-g9162: updated to !=

shared/api.ts
```diff
@@ -17,6 +18,27 @@ import {
 import {TelemetryConstants} from './telemetry/TelemetryConstants';
 import NetInfo from '@react-native-community/netinfo';
 
+let firstCall = true;
```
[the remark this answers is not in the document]
Sairam-g9162: this is to ensure that even the first cache that we do is also has TTL time as per config

abhip2565:
please add story details in description for tracking

swatigoel (changes requested):

shared/api.ts
```diff
@@ -264,10 +283,17 @@ async function generateCacheAPIFunctionWithAPIPreference(
   fetchCall: (...props: any[]) => any,
   onErrorHardCodedValue?: any,
 ) {
+  let cacheObject;
   try {
     const response = await fetchCall();
-    setItem(cacheKey, response, '').then(() =>
-      console.log('Cached response for ' + cacheKey),
+    if (response) {
```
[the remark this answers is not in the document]
abhip2565: Just thinking, should we handle this at `request.ts` level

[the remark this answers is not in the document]
swatigoel: @abhip2565 It should be part of api response handling
Api will give either response or error. Earlier also we were caching api response here as we have defined cache preferred and api preferred functions to handle cache. So let it be here only but handle response undefined scenario.
`request.ts` doesn't deal with looking into cache if response not found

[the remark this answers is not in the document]
abhip2565: Okay.. not cache part, only the response is null/emty/undefined scenario.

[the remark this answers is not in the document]
swatigoel: okay, response is null/empty/undefined from api response or cache response that need to be handle here right?

machines/app.ts
```diff
@@ -152,11 +154,32 @@ export const appMachine = model.createMachine(
...
+                {
+                  actions: [
+                    (_, event) =>
+                      console.info(
+                        'Error in fetching config, using hardcoded cacheTTL',
+                        JSON.stringify(event, null, 4),
+                      ),
+                  ],
+                },
+              ],
```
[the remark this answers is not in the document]
Sairam-g9162: Updated accordingly so even the config fetch failure will not disturb the app state

PR #1957, opened 2025-06-10
[INJIMOB-3242]: update logic to create detached JWS for OVP flow

Alka1703:


swatigoel:
Not needed as handled here https://github.com/mosip/inji-wallet/pull/1960

PR #1958, opened 2025-06-11
[INJIMOB-3242] Use jsonwebsignature2020 for signing VP with ED25519Signature2020 algo

Alka1703:


machines/openID4VP/openID4VPServices.ts
```diff
@@ -46,33 +50,40 @@ export const openID4VPServices = () => {
...
-            credentials,
-            context.keyType,
-          );
+          if (isIOS()) {
+            const canonicalizedDataToSign = await canonicalize(
+              JSON.parse(credentials.dataToSign),
+            );
+            proof = await constructDetachedJWT(
+              context.privateKey,
+              canonicalizedDataToSign ?? '',
```
swatigoel, line 72: shouldn't we throw error if data is not canonicalized instead of defaulting it to empty?

PR #2008, opened 2025-07-14
Enabling Mock VCs Verifiable

Sairam-g9162:
Enabling the Mock Vcs verifiable by removing the bypass we had earlier for the MockVC


shared/openId4VCI/Utils.ts
```diff
@@ -470,21 +470,12 @@ export function getMatchingCredentialIssuerMetadata(
 export async function verifyCredentialData(
   credential: Credential,
   credentialFormat: string,
-  issuerId: string,
 ) {
-  if (credentialFormat === VCFormat.mso_mdoc || !isMockVC(issuerId)) {
```
swatigoel, line 475: `isMockVC` usage is removed, why not import?
We can remove `isMockVC` method from Utils.

swatigoel (changes requested):

PR #2080, opened 2025-09-15
[INJIMOB-956]: SVG rendering

balachandarg-tw:
- If renderMethod present in VC, send the VC , wellknown and format to library and library returns the svg template in which placeholders are replaced with VC json. (labels are replaced with wellknown)
- If renderMethod is not present, fallback to wallknown.
- If QR code is present in the SVg Tempalte, show the `Tap to Zoom Qr code` button and allow user to share Qr code.

### DIFF
--- a/.talismanrc
+++ b/.talismanrc
@@ -341,7 +341,7 @@ fileignoreconfig:
   - filename: android/app/src/main/java/io/mosip/residentapp/InjiVciClientModule.java
     checksum: 17f55840bab193bc353034445ba4fce53e1ce466e95f616c15a1351f8d2f23bc
   - filename: ios/Inji.xcworkspace/xcshareddata/swiftpm/Package.resolved
-    checksum: 190ea27d8f86f16070ee381c28cf1b671d52c0bc1a18950bf7c9a8c8bfde4d31
+    checksum: 252427dd3d91cc71d644c0448e07683e3b7f3ff3936b31b075691c1f1cd0ba0d
   - filename: injitest/src/main/resources/Vids.json
     checksum: 8bcffed7a6dd565ae695e1b29de0655e10bd5c5420af2718defd593a687b8817
   - filename: injitest/src/main/java/inji/utils/UpdateNetworkSettings.java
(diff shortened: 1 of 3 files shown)


ios/RNInjiVcRenderer.m
```diff
@@ -0,0 +1,14 @@
+#import <Foundation/Foundation.h>
+#import "React/RCTBridgeModule.h"
+
+@interface RCT_EXTERN_MODULE(InjiVcRenderer, NSObject)
+
+RCT_EXTERN_METHOD(init:(NSString *)traceabilityId)
+
+RCT_EXTERN_METHOD(renderVC:(NSString *)vcJsonString
```
swatigoel, line 8: should we keep contract consistent?
first parameter is credentialFormt, then wellknown and last vcjson

shared/vcRenderer/VcRenderer.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (cachedRaw && typeof cachedRaw === 'string') {
+      try {
+        const cached: CachedSvg = JSON.parse(cachedRaw);
+        const withinTTL = Date.now() - cached.timestamp < CACHE_TTL;
+
+        if (withinTTL) {
+          return cached.data;
+        }
+      } catch (e) {
+        console.warn('::::failed to parse cached SVG, ignoring::::', e);
```
swatigoel, line 51: should we clear cache here?

shared/vcRenderer/VcRenderer.ts
```diff
@@ -0,0 +1,82 @@
...
+    vcJson: string,
+  ): Promise<string[]> {
+    const vc = JSON.parse(vcJson);
+    const vcId = vc.id ?? JSON.stringify(vc);
+    const cacheKey = this.createCacheKey(vcId);
+    const cachedRaw = await MMKV.getItem(cacheKey);
+    if (cachedRaw && typeof cachedRaw === 'string') {
+      try {
+        const cached: CachedSvg = JSON.parse(cachedRaw);
+        const withinTTL = Date.now() - cached.timestamp < CACHE_TTL;
```
swatigoel, line 45: should we try to have utility method which can be used across app?
It can be isCacheExpired. Passed timestamp of cache to it and it will take care of difference between current time and compare with TTL

ios/RNInjiVcRenderer.m
balachandarg-tw: Updated

shared/vcRenderer/VcRenderer.ts
```diff
@@ -0,0 +1,82 @@
...
+
+        if (withinTTL) {
+          return cached.data;
+        }
```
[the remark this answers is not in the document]
balachandarg-tw: Done

PR #2250, opened 2026-03-09
INJIMOB-3780 - feat: enable support for rendering jwt_vc

KVD-1302:
## Description
**Overview**
This PR enables end-to-end support for the jwt_vc_json credential format within Inji Wallet. It includes the necessary changes to register the format, process the JWT payload for UI rendering, and extract metadata fields from the issuer's well-known configuration.

**Changes Made**

- Format Registration: Added jwt_vc_json to the VCFormat enum to allow the system to recognize and route this new credential type.
- Payload Processing: Implemented a new processing block in VCProcessor.ts that utilizes jwt-decode to extract the credentialSubject and W3C type array from the raw JWT.
- UI Integration: Updated Utils.ts to trigger the VCProcessor during the rendering lifecycle when the jwt_vc_json format is detected.
- Metadata Extraction: Enhanced metadata logic to pull available fields from the issuer's .well-known configuration. This allows the wallet to dynamically map fields (like employeeId, name, role) for display.

Verification Status
Rendering: Successfully verified that jwt_vc_json credentials display correctly on the Home Screen and within the ID Details view.

[INJIMOB-3780](https://mosip.atlassian.net/browse/INJIMOB-3780)

## Screenshots
image
image

## Video
https://github.com/user-attachments/assets/6fe914f7-c088-4939-b7e2-23c283dfa75f

https://github.com/user-attachments/assets/1c4a9dbf-c7b7-4071-b701-b7c24ddefab5

## Summary by CodeRabbit

* **New Features**
  * Added support for a JWT-based VC JSON credential format so the app can decode JWT credentials and display their contents.
  * Field extraction and credential rendering now treat this JWT VC JSON format like existing SD-JWT formats, improving display of credential subject fields and labels.

* **Tests**
  * Updated tests to include the new format and adjusted format count expectations.


shared/openId4VCI/Utils.ts
```diff
@@ -84,6 +84,12 @@ export const updateCredentialInformation = async (
       context.selectedCredentialType.format,
     );
   }
+  if (context.selectedCredentialType.format === VCFormat.jwt_vc_json) {
```
KiruthikaJeyashankar: Issue (code smell):
Duplicate conditional branches all executing the same logic

Fix:
Combine the conditions (vc format = mso_mdoc / vc_sd_jwt / dc_sd_jwt/ jwt_vc_json) and call the method once.

KiruthikaJeyashankar (changes requested):

components/VC/common/VCProcessor.ts
```diff
@@ -38,6 +38,19 @@ export class VCProcessor {
         pathToDisclosures,
       };
     }
+    if (vcFormat === VCFormat.jwt_vc_json) {
+      const rawJwt = vcData.credential.toString();
+      const payload: any = jwtDecode(rawJwt);
+      const credentialSubject = payload.vc?.credentialSubject ?? {};
```
abhip2565: do we need `fullResolvedPayload` or `rawCredential` later anywhere? Why are we populating the object with these fields?

shared/openId4VCI/Utils.ts
```diff
@@ -190,6 +186,16 @@ export const getCredentialIssuersWellKnownConfig = async (
...
           }
+        } else if (format === VCFormat.jwt_vc_json) {
+          const jwtVcJsonFields = Object.keys(
+            matchingWellknownDetails.credential_definition?.credentialSubject ||
+              {},
+          );
+
+          if (jwtVcJsonFields.length > 0) {
+            fields = jwtVcJsonFields;
+            wellknownFieldsFlag = true;
```
abhip2565, line 197: VCUtils.tsx is the primary file for rendering, why are we not making any changes there?
https://github.com/inji/inji-wallet/blob/a62efc0bea102dbdce94def568e1cb3ac770bce9/components/VC/common/VCUtils.tsx#L74
http://github.com/inji/inji-wallet/blob/a62efc0bea102dbdce94def568e1cb3ac770bce9/components/VC/common/VCUtils.tsx#L178

abhip2565 (changes requested):

components/VC/common/VCProcessor.ts
KVD-1302: Fields like rawCredential, issuer, and type were added during the initial local integration testing to debug the data flow and are not actually required by the existing components.

I have refactored the jwt_vc_json block in VCProcessor.ts to return only the necessary fields:

fullResolvedPayload and publicKeys: Kept to maintain consistency with the SD-JWT pattern used in this file and ensure the existing rendering logic in VCUtils.tsx and VCDetailView.tsx works correctly.

credentialSubject: Included as the primary source of claims for display.

shared/openId4VCI/Utils.ts
KVD-1302: I have now explicitly integrated jwt_vc_json into the rendering logic within VCUtils.tsx.

While the UI initially rendered correctly earlier, it happened via an automatic fallback. I've added the format to the getFieldValue, getFieldName, and getFaceAttribute logic blocks. Because jwt_vc_json shares the same underlying payload structure as sd_jwt, grouping them ensures full support for nested claims and localized field names while maintaining a clean, DRY implementation. verified that all ID details render correctly with these explicit mappings.

components/VC/common/VCProcessor.ts
```diff
@@ -38,6 +38,16 @@ export class VCProcessor {
...
+      return {
+        fullResolvedPayload: credentialSubject,
+        publicKeys: Object.keys(credentialSubject),
+        credentialSubject,
```
KiruthikaJeyashankar: @KVD-1302 Wanted to understand on why the same - credentialSubject is being help in 2 key names?

components/VC/common/VCProcessor.ts
```diff
@@ -38,6 +38,16 @@ export class VCProcessor {
...
+      };
```
swatigoel, line 48: Why we need `publicKeys` here?
jwt vc has all the claims inside vc.credentialSubject. We just need those to render the VC

KiruthikaJeyashankar (changes requested):

components/VC/common/VCProcessor.ts
KVD-1302: You're right, publicKeys was a carryover from the sdjwt logic. I wanted to add it initially to maintain consistency but it is not needed here. I've removed it. I kept fullResolvedPayload as it's the primary key our rendering logic currently expects.

components/VC/common/VCProcessor.ts
KVD-1302: Cleaned this up! I removed the redundant keys so we're just returning fullResolvedPayload. It keeps things minimalistic while staying compatible with the existing UI components.

KiruthikaJeyashankar (approved):

components/VC/common/VCUtils.tsx
```diff
@@ -215,7 +216,11 @@ export const getFieldName = (
           return fieldName;
         }
       }
-    } else if (format === VCFormat.vc_sd_jwt || format === VCFormat.dc_sd_jwt) {
+    } else if (
+      format === VCFormat.vc_sd_jwt ||
+      format === VCFormat.dc_sd_jwt ||
+      format === VCFormat.jwt_vc_json
```
KiruthikaJeyashankar: wellknown of jwt_vc_json is different from that of other format, please handle accordingly

KiruthikaJeyashankar (changes requested):

components/VC/common/VCUtils.tsx
KVD-1302: I have refactored VCUtils.tsx to handle jwt_vc_json independently for both field name resolution and value extraction.

Metadata Isolation: Labels for jwt_vc_json are now resolved strictly from the credential_definition.credentialSubject path as per the Draft 13 specification, while SD-JWT remains on the claims path.

Independent Parsing: Rendering and payload access for JSON-based VCs are now handled in dedicated blocks to prevent any cross-format logic overlaps.

components/VC/common/VCUtils.tsx
```diff
@@ -236,6 +250,54 @@ export const getFieldName = (
         return getLocalizedField(newFieldObj);
       }
 
+      return formatKeyLabel(pathParts[pathParts.length - 1]);
+    } else if (format === VCFormat.jwt_vc_json) {
+      const pathParts = field.split('.');
+      const config =
```
KiruthikaJeyashankar: Do we need the statement?
```
 wellknown.credential_configurations_supported?[  'JwtVerifiableCredential'] 
```

Would just the below line workout?
```
const credentialSubject = wellknown.credential_definition?.credentialSubject;
```

Asking since the wellknown parameter in this function will already the value of wellknown.credential_configurations_supported?.['JwtVerifiableCredential']

components/VC/common/VCUtils.tsx
```diff
@@ -236,6 +250,54 @@ export const getFieldName = (
...
+        }
+
+        const newFieldObj: LocalizedEntry[] = currentObj.display.map(
+          (obj: any) => ({
+            language: obj.locale,
+            value: obj.name,
+          }),
+        );
+
+        const currentAppLang = i18n.language;
```
KiruthikaJeyashankar: getting the Field name's relevant display value as per app language is already handled in getLocalizedField, do we need this check here again?

components/VC/common/VCUtils.tsx
```diff
@@ -236,6 +250,54 @@ export const getFieldName = (
...
+            entry.language === currentAppLang ||
+            currentAppLang.startsWith(entry.language) ||
+            entry.language.startsWith(currentAppLang.split('-')[0]),
+        );
+        if (matchedEntry) {
+          return matchedEntry.value;
+        }
+
+        const localized = getLocalizedField(newFieldObj);
+        if (
```
KiruthikaJeyashankar: Do we need this check of if localized field not available then take the first entry of the display?
Its already handled in getLocalizedField

KiruthikaJeyashankar: follow this pattern 

```
if(fieldObject.display) {
 // get the localized field
}
// return the field name itself
```

ref - https://github.com/inji/inji-wallet/pull/2250/changes#diff-26ac40bd7004bb299ea1e34a072ac1ede192f28c25e316c0ca9f5a416fcf6e32R207-R214

KiruthikaJeyashankar (changes requested):

components/VC/common/VCUtils.tsx
KVD-1302: I'll simplify it to use wellknown.credential_definition?.credentialSubject directly as you suggested. Thank you.

components/VC/common/VCUtils.tsx
KVD-1302: While I’ve followed the suggested pattern and the primary call to getLocalizedField, I found that the helper doesn't support locale sub-tag matching (like hi-IN to hi). This caused the UI to default to the wrong language entry.

I’ve re-introduced a small, targeted fuzzy match to bridge this gap and ensure Hindi/English render correctly for the demo, while still using the global helper as the final fallback.

components/VC/common/VCUtils.tsx
KVD-1302: I tried relying solely on getLocalizedField, but I found that it doesn't handle locale sub-tag matching for certain Android devices.

For example, if the app reports hi-IN but the issuer metadata only provides a generic hi key, the helper fails to find a match and defaults to the first entry in the array (often English). I've re-added this targeted fuzzy check to ensure the UI remains robust for these common locale variations while still using the global helper as the primary fallback.

components/VC/common/VCUtils.tsx
```diff
@@ -490,12 +534,38 @@ export const fieldItemIterator = (
...
+  const disclosedKeys =
+    (verifiableCredential as {disclosedKeys?: string[]}).disclosedKeys || [];
   const renderedFields = new Set<string>();
 
-  const renderedMainFields = fields.map(field => {
+  const sortedFields = [...fields].sort((a, b) => {
+    if (format === VCFormat.jwt_vc_json) {
+      const config =
+        wellknown.credential_configurations_supported?.[
+          'JwtVerifiableCredential'
```
KiruthikaJeyashankar: @KVD-1302 credential configuration ID is a dynamic property - Keeping it hardcoded for JwtVerifiableCredential would not workout for places where credential configuration ID would be different.
Is it mandatory to have this hardcoded `JwtVerifiableCredential` here?

components/VC/common/VCUtils.tsx
```diff
@@ -490,12 +534,38 @@ export const fieldItemIterator = (
...
+
+        const orderA = getIndex(a);
+        const orderB = getIndex(b);
+        return (orderA === -1 ? 999 : orderA) - (orderB === -1 ? 999 : orderB);
+      }
+    }
+    return 0;
+  });
+
+  const renderedMainFields = sortedFields.map(field => {
```
KiruthikaJeyashankar: Wanted to understand the reason behind the change of fields to sorted fields?
Could you please clarify on this?

components/VC/common/VCUtils.tsx
KiruthikaJeyashankar: @KVD-1302 As of now, app supports 2 / 3 letter language code for rendering in UI. Shall we align with it?
If required for language support with subtags - it needs to handled for all formats of Vc which can be taken together later.

For now, can the localized field obtaining part be kept consistent?

KiruthikaJeyashankar (changes requested):

components/VC/common/VCUtils.tsx
KVD-1302: I see now in shared/openId4VCI/Utils.ts that fields is already being explicitly set to matchingWellknownDetails.order before it reaches fieldItemIterator. I was effectively re-sorting an already ordered array. I have removed the sortedFields block entirely and reverted back to fields.map. Thanks for pointing me to the exact file!

KiruthikaJeyashankar (approved):
LGTM

swatigoel (approved):

abhip2565:
lgtm

abhip2565 (approved):

PR #2405, opened 2026-04-24
fix: resolve ReferenceError in hasEncryptionKey service

### LINKED ISSUE #2404: ReferenceError in hasEncryptionKey
### Short Summary of the Bug

ReferenceError in `hasEncryptionKey` service due to undefined `event`

---

### Description

A `ReferenceError` exists in the `hasEncryptionKey` service within `machines/store.ts`. The code attempts to access `event.requester` when handling a `BiometricCancellationError`, but the `event` object is not defined within the scope of this callback service.

---

### Location

* **File**: `machines/store.ts`
* **Line**: 367

---

### Root Cause

The `hasEncryptionKey` service is defined as a simple async callback service that does not receive an `event` argument:

```typescript
hasEncryptionKey: () => async (callback) => {
  // ... code ...
  } catch (e) {
    if (e instanceof BiometricCancellationError) {
      // 'event' is undefined here
      callback(model.events.BIOMETRIC_CANCELLED(event.requester)); 
    }
  }
}
```

Unlike the `store` service defined later in the same file, which receives events via `onReceive`, `hasEncryptionKey` is invoked during a state transition and does not have access to a triggering event object.

---

### Impact

If a user cancels a biometric authentication prompt while the app is in the `checkEncryptionKey` state (part of the startup/initialization flow), the application will crash with:

```
ReferenceError: event is not defined
```

This prevents graceful handling of biometric cancellations and results in a poor user experience or app hang.

---

### Steps to Reproduce

1. Launch the app on a device/emulator where biometric auth is required for storage access
2. Reach the state where the app invokes `hasEncryptionKey` (e.g., during fresh install check)
3. When the biometric prompt appears, click **Cancel**
4. Observe the runtime crash or check logs for `ReferenceError`

---

### Attachments / Evidence / Links

```typescript
hasEncryptionKey: () => async (callback) => {
  ...
  } catch (e) {
    if (e instanceof BiometricCancellationError) {
      callback(model.events.BIOMETRIC_CANCELLED(event.requester)); // <--- ERROR HERE
    }
  }
}
```

CodeBySayak:
## Summary
This PR fixes a `ReferenceError` in the `hasEncryptionKey` service within `machines/store.ts`. The code was attempting to access `event.requester` when handling a `BiometricCancellationError`, but the `event` object is not defined within the scope of this callback service.

## Changes
- **machines/store.ts**: Removed the usage of `event.requester` in the `hasEncryptionKey` callback.
- Called `model.events.BIOMETRIC_CANCELLED()` without arguments to safely handle biometric cancellation.

## Impact
This fix prevents an application crash during the startup/initialization flow. Specifically, it addresses the scenario where a user cancels a biometric authentication prompt while the app is checking for an encryption key. Previously, this cancellation would trigger a `ReferenceError: event is not defined`, leading to an unstable app state.

## Testing
- Verified that the `BIOMETRIC_CANCELLED` event model supports an optional requester.
- Confirmed that removing the reference prevents the `ReferenceError` in the async callback scope.

closes #2404 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved biometric cancellation error handling to prevent unreliable event processing.


swatigoel:
@CodeBySayak can you raise PR for develop branch and sign off the commit too?

PR #2407, opened 2026-04-25
fix: atomic VC data+meta write to prevent credential loss on crash

Sanchit2662:
## Description

While digging through the credential issuance flow, I found a bug that silently deletes a credential right after it's downloaded if the app gets killed at the wrong moment.
                                                                                                                                                                                   
 The wallet saves two things when a credential is stored: the VC file to the filesystem, and a metadata pointer to MMKV. These were fired as two independent async messages ; metadata first, VC data second. MMKV is near-instant, so the pointer would land on disk before the file finished writing. If Android killed the app in that gap, the next launch would find a metadata entry with no corresponding file, classify it as tampered, and silently delete it. No error shown , the credential just disappears.                        
  
The fix was to combine both writes into a single atomic STORE_VC_WITH_META event handled by the store machine. It awaits the VC file write first, then writes the metadata  pointer. Nothing can interleave between the two. A crash before the file write means nothing is stored , the user retries. A crash between the two writes leaves a harmless orphan file. Only after both succeed does the credential appear in the wallet.                                                                                                   
  
  I also fixed two issues in the same state: the machine would hang forever if the user wasn't signed into cloud backup (a missing fallback in onDone), and there was no error handler if the backup check itself failed.
                                            

## Release Notes

* **Bug Fixes**
  * Improved error handling for credential processing operations with better error messaging and recovery paths
  * Optimized credential storage workflow to ensure proper data sequencing and consistency
  * Enhanced authentication state detection to prevent unnecessary re-authentication flows


abhip2565 (changes requested):
The objective mentioned in the description — atomic VC data+meta write — isn't fully served by this change. A few concerns:

- **Not atomic:** `setItem` and `prependItem` are still two separate awaits. A crash between them leaves an orphan file with no metadata pointer.
- **No rollback:** If the metadata write fails after the VC file write succeeds, the orphan file isn't cleaned up.
- **Still sequential:** This effectively reverses the write order (VC file first, metadata second instead of the other way around), which makes the failure mode less destructive — but doesn't eliminate it. The gap between the two writes still exists.

The `isUserSignedAlready` fallback and error handler additions are good catches though.

Sanchit2662:
You're right @abhip2565  that it's not truly atomic at the syscall level , can't be, since MMKV and FileStorage don't share a transaction coordinator.
What I did instead is a WAL: before touching either storage backend, I write the vcKey to a journal in MMKV. Then VC file, then metadata, then clear the journal. On every cold start, reconcilePendingVcWrites runs before credentials are loaded and heals whatever was left half-done , if the file got written but metadata didn't, it completes it; if neither got written, it just drops the journal entry. Every crash window has a recovery path, so nothing ends up permanently inconsistent from the user's side.

abhip2565:
Thinking about this more — the simplest fix for the credential loss bug is just swapping the order of `storeVerifiableCredentialData` before `storeVerifiableCredentialMeta` in `storing.entry`. That's a 2-line change.

The metadata list is the index; the UI loads credentials by iterating it. If the file writes first and the app dies before metadata lands, we get an orphan file that's invisible to the user (rare scenario as Mmkv is very fast)— no pointer means it's never loaded, never flagged as tampered, never deleted. The orphan sits on disk harmlessly and the user retries issuance. If orphan accumulation becomes a concern, a simple startup sweep (FS files with no metadata entry → delete) handles it.

The new event type, the journal, and the reconciler all add surface area for a crash window that's already benign once the order is flipped. I'd suggest scoping this PR down to:

1. Swap the two entry actions (eliminates the destructive path)
2. Keep the `isUserSignedAlready` fallback and `processingCredential.onError` (real bug fixes)
3. Drop `storeVCWithMeta`, the journal, and `reconcilePendingVcWrites`

We can revisit durability guarantees separately if there's a concrete need.

Sanchit2662:
Fair, you're right @abhip2565  once the order's flipped the orphan-file case is benign and the journal is paying for a problem that doesn't exist anymore. Scoped down to: swapped entry order, the isUserSignedAlready fallback, and proccessingCredential.onError. Dropped the new event, the journal, and the reconciler. We can revisit durability if it ever bites   us.

PR #2409, opened 2026-04-26
fix: add missing target to openID4VP onError handlers to prevent deadlock

Sanchit2662:
## Description

I was tracing through the OpenID4VP flow when I noticed three onError handlers , on getTrustedVerifiersList, getKeyPairFromKeystore, and checkKeyPair , had no target defined. In XState, that means when the invoked service rejects, the machine just... sits there. No transition, no error state, no recovery. The UI freezes and the user has no way out except killing the app.                                                                                                     
This hits in real conditions: a dropped network call while fetching the trusted verifiers list, or a biometric prompt that gets cancelled or times out. Both are normal user behavior.                                                                                                                                             
The fix is straightforward , I added target transitions to all three handlers, using the same branching pattern already used by authenticateVerifier.onError in   the same file: route to #OpenID4VP.authFlowFailed when it's an authorization flow, and to showError otherwise. The showError state already has a RETRY handler, so users get a proper recovery path instead of a frozen screen.                                                                                                 
  
No new logic, no new states. Just wiring up the exits that were always supposed to be there.


Sanchit2662:
Hey @abhip2565 , please review the fix.

abhip2565 (approved):

abhip2565:
please attach a sanity video of the fix, and happy path - vc download.

Sanchit2662:
Hey @abhip2565, wanted to give you an update on the sanity video.

So I tried running the app to record the happy path but ran into a blocker - the QA backend (api.qa-inji1.mosip.net) is completely unreachable right now. I switched to the collab environment (api.collab.mosip.net) and rebuilt the APK with that, but even then the app throws a network error when trying to load issuers.

I tested the collab URLs directly and they return 200, so the server is up. The issue seems to be that the device I'm testing on can't reach it for some reason, maybe a firewall or network restriction on my end.

The fix itself is solid though - three onError handlers that were missing target transitions, which would've left the machine stuck with no way out. The pattern I used is exactly the same as authenticateVerifier.onError already in the file. Nothing new, just wiring up what was always supposed to be there.

Happy to answer anything about the fix if needed.

swatigoel:
@Sanchit2662 OpenId4VP feature is enhanced by adding support for DCQL, hence major refactoring happened. After that refactoring this issue is no more valid. 
Please refer this https://github.com/inji/inji-wallet/blob/develop/machines/openID4VP/openID4VPMachine.ts

PR #2422, opened 2026-04-27
fix(app): Update credentail Registry for  reduce launch delay

Kaushikgupta469:
## INJI-Mobile app startup fix

> Older credential registry was decomissoned resulting in app trying to fetch it and fallback to cached value which caused the app delay in startup

> Example : [Issue-2400](https://github.com/inji/inji-wallet/issues/2400)

## Summary by CodeRabbit

* **Chores**
  * Updated API endpoint configuration to point to the correct MOSIP service endpoint.

### DIFF
--- a/.env
+++ b/.env
@@ -2,7 +2,7 @@
 # eg . npm build android:newlogic --reset-cache
 
 #MIMOTO_HOST=http://mock.mimoto.newlogic.dev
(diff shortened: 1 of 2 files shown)


swatigoel:
@Kaushikgupta469 we need to handle error gracefully in case mimoto is not reachable. Current solution will not fix the actual issue.

PR #2433, opened 2026-05-02
fix: prevent crash in AuthWebViewScreen when params are missing or invalid

synakr:
#### Issue

`route.params` is destructured and used before validation. If it is undefined or contains invalid values, it can cause runtime crashes (e.g., during URL parsing with `new URL()`).

#### Changes

* Made params destructuring null-safe:

  ```ts
  const { authorizationURL, clientId, redirectUri, controller } = route.params ?? {};
  ```
* Guarded URL parsing to prevent runtime errors:

  ```ts
  let hostName = 'mosip.net';
  if (authorizationURL) {
    try {
      hostName = new URL(authorizationURL).hostname;
    } catch (err) {
      console.warn('Invalid authorizationURL:', authorizationURL);
    }
  }
  ```
* Extended validation to include all required parameters:

  ```ts
  if (!authorizationURL || !clientId || !redirectUri || !controller)
  ```
* Added safety check before redirect handling:

  ```ts
  if (redirectUri && url.startsWith(redirectUri))
  ```

#### Impact

Prevents crashes when navigation parameters are missing or malformed and ensures safe URL handling, without affecting existing behavior for valid inputs.

## Release Notes

* **Bug Fixes**
  * Improved robustness of OAuth authentication by implementing safer URL parsing with fallback handling
  * Enhanced redirect validation during authentication flows to prevent edge cases with missing or invalid parameters


swatigoel:
@synakr please check coderabbit comment and raise PR for develop

Issue #2400, opened 2026-04-23
Application launch is taking time ~1min


hegdenitin:
**Steps to Re-Produce :**
1. Install Application on Android/iOS devices
2. Open Inji wallet

 
Actual: Application is taking almost a minute launch. Till then there is white screen appears.
Expected: Application should launch without any delay

Video:
Android -
https://github.com/user-attachments/assets/572892d2-a368-4bce-ac09-deb3093d8491

iOS-

https://github.com/user-attachments/assets/fd61de81-7158-403b-941d-95da1a8c2355

Logs:
Android - 

[realme-RMX3371-Android-14_2026-04-23_101910.log](https://github.com/user-attachments/files/26998050/realme-RMX3371-Android-14_2026-04-23_101910.log)

iOS - 

[iPhone13_iOS_24.6.1_Logs_forAppLaunch.log](https://github.com/user-attachments/files/26998058/iPhone13_iOS_24.6.1_Logs_forAppLaunch.log)

Phone make/model: Realme GT NEO 3T
OS: Android 14
BLE version : 5.2
Applications tested with: 238.apk (0.23.0)

Phone make/model: iPhone 13
OS: iOS 26.4.1
BLE version : 5.2
Applications tested with: 0.23.0 (9)

PR #2453, opened 2026-05-19
fix: application launch delay logic for wallet app 

Kaushikgupta469:
## Description

>Application launch delay Fixes issue #2400 

## Summary by CodeRabbit

* **Performance**
  * Cache operations made non-blocking to improve responsiveness; configuration fetch now uses a 10,000 ms timeout.

* **Bug Fixes**
  * More reliable fallback to cached content on failures; reduced noisy errors logged and suppressed non-critical failures.

* **Behavior**
  * Offline/network gating removed for cache reads so stale cached data is served more readily when fetches fail.

### DIFF
--- a/shared/api.ts
+++ b/shared/api.ts
@@ -267,6 +267,9 @@ async function generateCacheAPIFunctionWithCachePreference(
     console.log(error);
 
     if (onErrorHardCodedValue != undefined) {
+      setItem(cacheKey, createCacheObject(onErrorHardCodedValue), '').catch(
+        () => {},
+      );
       return onErrorHardCodedValue;
     } else {
       throw error;
@@ -279,7 +282,30 @@ async function generateCacheAPIFunctionWithAPIPreference(
   fetchCall: (...props: any[]) => any,
   onErrorHardCodedValue?: any,
 ) {
-  let cacheObject;
+  const existingCache = await getItem(cacheKey, null, '');
+  if (existingCache && isCacheValid(existingCache)) {
+    const refreshInBackground = async () => {
+      try {
+        const freshResponse = await fetchCall();
+        if (!freshResponse) return;
+        if (freshResponse?.cacheTTLInMilliSeconds) {
+          updateCacheTTL(Number(freshResponse.cacheTTLInMilliSeconds));
+        }
+        setItem(cacheKey, createCacheObject(freshResponse), '').then(() =>
+          console.info('Cached response for ' + cacheKey),
+        );
+      } catch (refreshError) {
+        console.debug(
+          'Background cache refresh failed for',
+          cacheKey,
+          refreshError,
+        );
+      }
+    };
+    refreshInBackground();
+    return existingCache.response;
+  }
+
   try {
     const response = await fetchCall();
     if (!response) {
@@ -288,8 +314,7 @@ async function generateCacheAPIFunctionWithAPIPreference(
     if (response?.cacheTTLInMilliSeconds) {
       updateCacheTTL(Number(response.cacheTTLInMilliSeconds));
     }
-    cacheObject = createCacheObject(response);
-    setItem(cacheKey, cacheObject, '').then(() =>
+    setItem(cacheKey, createCacheObject(response), '').then(() =>
       console.info('Cached response for ' + cacheKey),
     );
     return response;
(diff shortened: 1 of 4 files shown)


shared/api.ts
```diff
@@ -268,6 +267,14 @@ async function generateCacheAPIFunctionWithCachePreference(
     console.log(error);
 
     if (onErrorHardCodedValue != undefined) {
+      setItem(cacheKey, createCacheObject(onErrorHardCodedValue), '').catch(
```
KiruthikaJeyashankar: Why is hardcoded value put into cache?
- If a cache preferred API is called - even if the actual API result is reachable - cache is preferred
- Given this is the case the env went down for sometime - hardCodedValue is populated as cache - now if the env is up again - the cached hardcoded value is available still cache expiry
Will this scenario be okay?

@coderabbitai checkout on this case

shared/api.ts
```diff
@@ -280,7 +287,30 @@ async function generateCacheAPIFunctionWithAPIPreference(
   fetchCall: (...props: any[]) => any,
   onErrorHardCodedValue?: any,
 ) {
-  let cacheObject;
+  const existingCache = await getItem(cacheKey, null, '');
```
KiruthikaJeyashankar: @Kaushikgupta469 Why is cache checked on **generateCacheAPIFunctionWithAPIPreference** function?

This operates on
1.Make api call
2. If API call fails then get from cache

shared/api.ts
```diff
@@ -300,13 +329,15 @@ async function generateCacheAPIFunctionWithAPIPreference(
...
     }`);
     console.error(`The error in fetching api ${cacheKey}`, error);
-    let cachedData = null;
-    if (!(await NetInfo.fetch()).isConnected) {
-      cachedData = await getItem(cacheKey, null, '');
-    }
-    if (cachedData && isCacheValid(cachedData)) {
-      return cachedData.response;
-    } else if (onErrorHardCodedValue != undefined) {
+    if (onErrorHardCodedValue != undefined) {
```
KiruthikaJeyashankar, line 315: If internet is not connected - cache was returned previously
why is that removed @Kaushikgupta469 ?

KiruthikaJeyashankar: Only if cache is not available - hardcoded value is returned

KiruthikaJeyashankar (changes requested):

KiruthikaJeyashankar:
@coderabbitai Context: This PR fixes https://github.com/inji/inji-wallet/issues/2400
Where Application launch is taking time ~1min when the env value is pointing to unavailable environment, then the getAllProperties which is called on app launch is taking more time to understand that the API is down
This eventually causes time delay here

Suggest relavent solution here

KiruthikaJeyashankar:
@coderabbitai The current (excluding changes of this PR) generateCacheAPIFunctionWithAPIPreference function - is it handling the case properly?

Case:
1. First make API call - if it succeeds - store in cache for later reference and return the response
2. API call fails ->  take from cache -> if cache fails take from hardcoded value

Kaushikgupta469:
@coderabbitai Please review the latest commit and verify whether all possible scenarios and edge cases are handled correctly. If anything is missing, incomplete, or could cause issues, please suggest the required fixes along with the relevant code changes.

shared/api.ts
```diff
@@ -243,29 +247,33 @@ async function generateCacheAPIFunctionWithCachePreference(
...
   } catch (error) {
     console.warn(`Failed to load due to network issue in cache preferred api call.
      cache key:${cacheKey} and has onErrorHardCodedValue:${
       onErrorHardCodedValue != undefined
     }`);
-    console.log(error);
+    console.warn(error);
+
+    if (existingCache?.response) {
+      console.info(`Returning stale cache for ${cacheKey}`);
```
KiruthikaJeyashankar, line 274: Should we return stale cache? is it okay to return?

shared/api.ts
KiruthikaJeyashankar: @Kaushikgupta469

KiruthikaJeyashankar (changes requested):

KiruthikaJeyashankar:
@Kaushikgupta469 Lets update .env to have MIMOTO_HOST pointed to collab env?

This way its easy for people trying out app but not having custom env
WDYT?

PR #2495, opened 2026-06-19
[KATTU-334] Update sonar workflow reference to develop

amanbh321:


Mahesh-Binayak (approved):

swatigoel:
not needed, as now we are using the tagged version

PR #2511, opened 2026-06-27
feat(dpop): attach DPoP proof to token request and handle nonce retry

abhip2565:
Wires the wallet side of DPoP (RFC 9449) per the DPoP ADR.

## Summary by CodeRabbit

* **New Features**
  * Added support for generating DPoP proofs for token requests.
  * Token requests now include DPoP proofs when available.
  * Automatically retries requests once with a refreshed proof when the issuer requests a new DPoP nonce.
  * Added support for proxy token endpoints.

* **Bug Fixes**
  * Improved token-request error handling and reporting for failed retries and missing endpoints.

* **Tests**
  * Added coverage for DPoP headers, nonce-based retries, failures, and proxy endpoints.

### DIFF
--- a/shared/openId4VCI/TokenService.ts
+++ b/shared/openId4VCI/TokenService.ts
@@ -58,9 +58,13 @@ function parseError(errorText: string): any {
 
 function toVciClientError(errorText: string): VciClientErrorResponse {
   const parsedError = parseError(errorText);
+  const issuerErrorCode = parsedError.error ?? 'UNKNOWN_ERROR';
+  const issuerErrorMessage = parsedError.error_description;
   return {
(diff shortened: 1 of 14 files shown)


shared/openId4VCI/TokenService.ts
```diff
@@ -0,0 +1,130 @@
+import VciClient, {VciClientErrorResponse} from '../vciClient/VciClient';
```
swatigoel, line 1: Instead of naming file as sendTokenRequest, we can name it TokenService / AccessTokenService or move this to https://github.com/inji/inji-wallet/blob/develop/shared/openId4VCI/Utils.ts

swatigoel (changes requested):

swatigoel (approved):

PR #2530, opened 2026-07-13
fix: show issuer contact message for HCERT verification failures

abhip2565:
## Summary
- Map HCERT QR decode and signature verification failures to a dedicated verification error code.
- Show the issuer-contact download failure message for that code in the loaded locale files.
- Keep the underlying HCERT failure details in logs for troubleshooting.

## Validation
- node -e "JSON.parse(require('fs').readFileSync('locales/en.json','utf8')); JSON.parse(require('fs').readFileSync('locales/es.json','utf8')); console.log('locale json ok')"
- ./node_modules/.bin/tsc --noEmit
- git diff --check


abhip2565:
Closing this PR because the changes should not be raised as a new pull request.

